PR #328, opened 2015-01-18
DOC: Improve error message

ElDeveloper:
It was pointed out that the error message printed when sample names didn't
match between coordinates and mapping file was unclear. So the message has
been rephrased.


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_328/make_emperor/

PR #338, opened 2015-02-03
DOC: Update link pointing to version 0.9.3

ElDeveloper:
The button in the documentation pointed to version 0.9.3 as the "stable" 
version, I've changed the link to point to the releases page in GitHub, which
should always list the latest release at the top.


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_338/make_emperor/

PR #347, opened 2015-02-22
BUG: Fix problem with Icon? files in OSX

### LINKED ISSUE #323: Compare plots fails if input directory contains files _other than_ pc.txt
For some reason, the transform coordinates file posted a "Icon?" file in the output directory, leading to the error:

if not lines[0].startswith('pc vector number'):
IndexError: list index out of range

Removing the non _pc.txt files solves this.

ElDeveloper:
make_emperor.py incorrectly tried to parse files named `Icon?` in OS X and
this was causing a non-informative error to come up on screen. A new function
to retrieve these filenames has been added along with a test_data folder to
exercise this function.

Fixes #323


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_347/make_emperor/

antgonza:
Conflicts!

ElDeveloper:
Conflicts should be resolved.

emperor/util.py
```diff
@@ -532,3 +532,45 @@ def sanitize_mapping_file(data, headers):
...
+    - Is folder
+    - Is part of the procrustes results from QIIME, see
+      transform_coordinate_matrices.py
+    """
+    coord_fps = []
+
+    for filepath in listdir(dir_path):
+        if filepath.startswith('.'):
+            continue
+        if filepath.startswith('Icon?') or filepath == 'Icon?':
```
antgonza: Shouldn't startswith catch the second parameter too? I tested on ipython and it does ...

ElDeveloper: Yes you are correct, I've fixed this, thanks!

emperor/util.py
```diff
@@ -532,3 +532,45 @@ def sanitize_mapping_file(data, headers):
...
+    ignored:
+    - Is a hidden file
+    - Is named `Icon?`.
+    - Is folder
+    - Is part of the procrustes results from QIIME, see
+      transform_coordinate_matrices.py
+    """
+    coord_fps = []
+
+    for filepath in listdir(dir_path):
```
antgonza: What about replacing this block with a glob.glob?

ElDeveloper: That probably makes sense, but since that would affect the `-i` option of `make_emperor.py` and this PR is only trying to fix one bug, would you mind opening a separate issue? I'm still not super convinced about the glob though, remember that we have to maintain compatibility with QIIME 1.9.0 and I'm worried that changing `-i` would break something.

antgonza: Sorry for not being clear. I was suggesting glob.glob so you could retrieve specific filenames from the file system and avoid looping on the files to find if they start with . or if they have x or y in the filename. I do not believe this is necessary but I was wondering if you have considered it.

ElDeveloper: Ahhh, now I know what you mean. I don't think we can use it, because we
need an "exclude" pattern which doesn't seem to be supported by python,
see:

```python
In [1]: from glob import glob

In [2]: glob('!(Do*)')
Out[2]: []

In [3]: ls 
Applications/
Desktop/
Documents/
Downloads/
Dropbox/
Library/
Mail/
Movies/
Music/
Pictures/
Public/
Rarefy_draft4.docx
Room andor AV Confirmation-53429.pdf
Screen Shot 2015-02-23 at 5.47.52 PM.png
Sites/
VirtualBox VMs/
git_sw/
hg_sw/
miniconda/
muttrc
svn_sw/
tmp/
workshop-schedule.pdf
```

PR #350, opened 2015-02-25
BUG: Fix bug with animations

ElDeveloper:
The framework would try to animate trajectories with a single unique 
timepoint, even though it makes no sense and those should be ignored as they
are not very relevant in the context of an animation.

I've added a new test case to check for this.

### DIFF
--- a/emperor/support_files/js/animate.js
+++ b/emperor/support_files/js/animate.js
@@ -162,8 +162,10 @@ AnimationDirector.prototype.initializeTrajectories = function(){
                               'z':chewedDataBuffer[index]['z']});
     }
 
-    // don't add a new trajectory
-    if (sampleNamesBuffer.length <= 1){
+    // Don't add a trajectory unless it has more than one different points
+    // in the gradient. There's no reason why we should animate a trajectory
+    // that has a timepoint in 0, 0 and 0 or one that has a timepoint in 0.
+    if (_.uniq(gradientPointsBuffer).length <= 1){
       continue;
     }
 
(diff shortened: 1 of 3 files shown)


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_350/make_emperor/

emperor/support_files/js/animate.js
```diff
@@ -162,8 +162,10 @@ AnimationDirector.prototype.initializeTrajectories = function(){
                               'z':chewedDataBuffer[index]['z']});
     }
 
-    // don't add a new trajectory
-    if (sampleNamesBuffer.length <= 1){
+    // Don't add a trajectory unless it has more than one different points
+    // in the gradient. There's no reason why we should animate a trajectory
+    // that has a timepoint in 0, 0 and 0 or one that has a timepoint in 0.
```
antgonza: Could you make this clearer?

tests/javascript_tests/test_animate.js
```diff
@@ -99,6 +105,21 @@ $(document).ready(function() {
                   {"x": -0.237661, "y": 0.046053, "z": -0.138136},
                   {"x": -0.276542, "y": -0.144964, "z": 0.066647}],
                   'Fast');
+        equal(director.trajectories.length, 2, 'The number of trajectories is '+
+              'correct');
+    });
+
+    test('Test useless trajectories are removed', function(){
```
antgonza: Could you also check that the number of points in the trajectories is correct? C:2, D:3

ElDeveloper: Good catch, thanks!

emperor/support_files/js/animate.js
```diff
@@ -162,8 +162,11 @@ AnimationDirector.prototype.initializeTrajectories = function(){
                               'z':chewedDataBuffer[index]['z']});
     }
 
-    // don't add a new trajectory
-    if (sampleNamesBuffer.length <= 1){
+    // Don't add a trajectory unless it has more than one samples in the
```
wasade: "...more than one sample..."

PR #354, opened 2015-03-05
ENH: Add coveralls 👖

ElDeveloper:
I wasn't able to get coverage to account for the script usage tests, but I
think this is fine for the time being.


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_354/make_emperor/

Issue #59, opened 2013-04-29
Allow filtering of taxa strings for biplots


walterst:
The full taxa strings can be cumbersome to visualize.

An example taxonomy string:
k__Bacteria;p__Firmicutes;c__Clostridia;o__Clostridiales;f__Ruminococcaceae;g__Ruminococcus;s__

The most informative (and least abusive of real estate) could have just g__Ruminococcus.

It might be helpful to allow the entire taxa string to be displayed, or one or more of the semicolon separated levels (so allow phylum and genus level to be displayed) as a column selected of the taxa.

Could also be handy to allow modification of the text (so only display genus level  g__Ruminococcus) and then delete the prefix, so it just shows Ruminococcus.

mortonjt:
I think I start looking into this problem

ElDeveloper:
AWESOME!! Thanks!!!

ElDeveloper:
This was fixed in #359

PR #359, opened 2015-03-08
Taxa label

mortonjt:
This PR is addressing issue https://github.com/biocore/emperor/issues/59

This will allow the user to toggle to truncate the taxa labels.  For instance, if the taxa label is 

`k__Bacteria; p__Firmicutes; c__;`

Only the following text will be displayed

`p__Firmicutes`


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_359/make_emperor/

ElDeveloper:
This is awesome, you came up with a really good and simple solution!!

Can you move the function `getLastKnownLevel` to either (1) emperor/support_files/js/util.js or (2) emperor/support_files/js/taxa.js, whichever you consider more appropriate. Once you do that, can you also add unit tests for this function.

For an example see [`test_animate.js`](https://github.com/biocore/emperor/blob/master/tests/javascript_tests/test_animate.js), if you notice, the structure is very similar to what you would see in a Python test file. We are using [QUnit](http://qunitjs.com), for the JavaScript unit testing. To run the tests you can use `all_tests.py` and use the `--suppress_unit_tests --suppress_script_usage_tests` to just run the JavaScript test suite, note that you are going to need [phantom.js](http://phantomjs.org) installed in your machine for this to work. Alternatively you can just open `tests/javascript_tests/index.html` in Chrome ****.

With that the only other thing missing would be adding a note to the ChangeLog.md file and this should be good to go.

-----

**** Note that there are two tests that fail in the browser but not in phantom.js so you can sefely ignore them if running in the browser, the tests are:

```
21 User Interface: Test color manager table construction with no extra parameters (1, 0, 1)Rerun
22 User Interface: Test color manager table construction with an extra parameter (1, 0, 1)Rerun
```

antgonza:
Really nice! Could you make another option so you display the last
string too? Some users might like this. For example: k__Bacteria;
p__Firmicutes; c__; can be:
- p__Firmicutes (what you have)
- p__Firmicutes; c__;

mortonjt:
I thought I put in a Toggle Taxa Label in there.  That should display the original taxa labels that we had before.  Is that not working for you @antgonza ?

antgonza:
Yes, saw that and I think I'm not being clear. Let me use another example:
1. k__Bacteria; p__Actinobacteria; c__Actinobacteria;
o__Actinomycetales; f__Streptomycetaceae; g__Streptomyces;
s__mirabilis
2. k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales;
f__Porphyromonadaceae; g__Paludibacter; s__

Current:
s__mirabilis
g__Paludibacter

Suggestion:
s__mirabilis
g__Paludibacter; s__

mortonjt:
So just display the last string to let the user know that it wasn't classified.  Good call!

ElDeveloper:
Can we have just one option though? Either one or the other would be
fine, but then again I barely ever use this feature so maybe you should
disregard my opinion, @walterst would you mind commenting on which one
is more useful?

If we end up needing two of them, then using a drop down menu would make
more sense. The options would be: No labels, Filtered String (what's
currently implemented) and Unfiltered string (what @antgonza is
proposing). What do you guys think?

antgonza:
Having 1 option is fine but I think the suggested will be better.

walterst:
I do like the suggested approach by Antonio a bit better than the single, last named, level of taxonomy, as it would be more informative. Perhaps even better, but maybe asking too much, is a sliding bar for the number of taxonomy levels to show (starting at 1 for the lowest level).

rob-knight:
Sliding bar would be _amazing_…

On Mar 9, 2015, at 9:49 AM, Tony <notifications@github.com<mailto:notifications@github.com>> wrote:


I do like the suggested approach by Antonio a bit better than the single, last named, level of taxonomy, as it would be more informative. Perhaps even better, but maybe asking too much, is a sliding bar for the number of taxonomy levels to show (starting at 1 for the lowest level).

antgonza:
Good idea! What about: http://blog.153.io/Elessar/?

ElDeveloper:
:+1: to slider! @mortonjt, let me know if you would like to talk about
the details to implement this.

mortonjt:
A sliding bar to display 1 level at a time? I like it!

mortonjt:
So something like [this?](http://jsfiddle.net/mortonjt/3ow8mcm3/1/)

antgonza:
That looks great, thanks! Note that the suggestion still stands of
showing the last known classification for those that do not have that
specific level.

ElDeveloper:
THAT'S AMAZING!!! :+1:

mortonjt:
@antgonza , let see if I understand what you are suggesting.

Say we only have two taxa

`k__Bacteria; p__Firmicutes`
`k__Bacteria; p__`

When the slider points to the phylum level, only the following should be displayed
`p__Firmicutes`
`p__`

Correct?

antgonza:
Nop, it should display the last known level:
p__Firmicutes
k__Bacteria; p__

mortonjt:
Gotcha.

Cool.  That won't be a problem.  Could clear up confusion down the road.

mortonjt:
Decided to use the sliding bars already present in Emperor - doesn't require anymore dependencies.

Still think we should upgrade to jQuery soon.  Things will probably break since jQuery isn't completely backwards compatible.  Will try to address in next PR.

emperor/support_files/js/emperor.js
```diff
@@ -870,17 +872,30 @@ function toggleLabels() {
...
+        var labelshtml = "";
+        $('#taxalabels').css('display','block');
+        $("#taxaLevel" ).html(g_taxonomy[g_taxaLevel]);
+        for(var key in g_taxaPositions){
+                // get the coordinate of this taxa sphere
+            	var coords = toScreenXY(g_plotTaxa[key].position,g_sceneCamera,$('#main-plot'));
+
+            	// labels are identified by the key they have in g_taxaPositions
+            	labelshtml += "<label id=\""+key+"_taxalabel\" class=\"unselectable labels\" style=\"position:absolute; left:"+parseInt(coords['x'])+"px; top:"+parseInt(coords['y'])+"px;\">";
+                labelshtml += truncateLevel(g_taxaPositions[key]['lineage'], g_taxaLevel);
```
ElDeveloper: I think you have an extra leading space in this line.

emperor/support_files/js/emperor.js
```diff
@@ -870,17 +872,30 @@ function toggleLabels() {
...
+                console.log("Lineage"+ g_taxaPositions[key]['lineage'])
```
ElDeveloper: Can you remove this line?

tests/test_core.py
```diff
@@ -36,7 +36,10 @@ def setUp(self):
 
     def test_str(self):
         emp = Emperor(self.ord_res, self.data, self.headers)
-        self.assertItemsEqual(HTML_STRING.split('\n'), str(emp).split('\n'))
+        #print(HTML_STRING.split('\n'))
```
ElDeveloper: Can you remove these lines?

emperor/support_files/js/emperor.js
```diff
@@ -48,7 +48,9 @@ var g_keyBuilt = false;
 var g_screenshotBind;
 var g_separator_left;
 var g_separator_history;
-
+var g_slider = null;
```
ElDeveloper: This variable seems to be unused.

emperor/support_files/js/emperor.js
```diff
@@ -48,7 +48,9 @@ var g_keyBuilt = false;
...
+var g_taxaLevel = 0;  // taxonomic level to display
```
ElDeveloper: Instead of using this variable to keep track of the selected level, I would retrieve the value directly from the slider.

emperor/support_files/js/emperor.js
```diff
@@ -48,7 +48,9 @@ var g_keyBuilt = false;
...
+var g_taxonomy = ["Full","Kingdom","Phylum","Class","Order","Family","Genus","Species"];
```
ElDeveloper: It doesn't seem like we need to have this as a global variable.

mortonjt:
Thanks for the hasty feedback!  Maybe I should squash these commits together?  Not only it would clean up the commits, but it could make debugging utilities like [git bisect](http://git-scm.com/docs/git-bisect) easier to use

ElDeveloper:
I generally support squashing! Does make the history cleaner and in a
lot of cases makes understanding a feature easier. That being said, I
will not require you that this is done for your PR to be merged as
historically we haven't been doing this.

ChangeLog.md
```diff
@@ -16,6 +16,7 @@ Emperor 0.9.51-dev (changes since Emperor 0.9.5 go here)
 
 ### New features
 
+* Slider bar added to display different levels of taxonomies
```
ElDeveloper: Can you make this line more descriptive? Something along the lines of:

```md
Add a slider bar under the `Labels` tab to select the taxonomic levels to display.
```

emperor/support_files/js/emperor.js
```diff
@@ -2261,25 +2274,23 @@ $(document).ready(function() {
...
-		}
-		document.getElementById("taxalabels").innerHTML = labelshtml
-
+                displayTaxaLabels(0);
+	        //add sliding bar to specify taxonomic labeling
+                $("#selectTaxaLevel").slider(
+                {
+                            value:0,
+                            min: 0,
+                            max: 7,
```
ElDeveloper: This number should only go up to the highest level that we have available in the taxa strings.

emperor/support_files/js/util.js
```diff
@@ -83,6 +82,34 @@ function naturalSort(list){
   return result.concat(alphaPart, numericPart);
 }
 
+
+/**
```
ElDeveloper: General comment about this, JavaScript uses camelcase instead of underscores, would you mind changing that for the variable names?

emperor/support_files/js/util.js
```diff
@@ -83,6 +82,34 @@ function naturalSort(list){
...
+ * Utility function that splits the lineage into taxonomic levels
+ * and returns the taxonomic level specified
+ *
+ * @param {node} XML DOM object, usually as created by the document object.
+ *
+ * @return string representation of the node object.
+ *
+*/
+function truncateLevel(lineage, level_index){
+    if (level_index==0){
```
ElDeveloper: Can you do triple equals?

emperor/support_files/js/util.js
```diff
@@ -83,6 +82,34 @@ function naturalSort(list){
...
+	return lineage;
+    }
+    var levels = lineage.split(';');
+    var taxa_label = '';
+    for( var i = 0; (i<level_index && i<levels.length); i++){
+        var level = levels[i];
```
ElDeveloper: The indentation of this for block is a bit funny, would you mind fixing that and making sure it is all two-spaced?

ElDeveloper:
I'm encountering a problem with taxa label coloring and the slider (points explained below the animated gif):

pr

- The background of the slider is filled in blue regardless of where the position of the slider is, this is inconsistent with other sliders in the interface.
- The colors of the labels are reseting whenever I move the slider.

Looks good, thanks for the work on this! 0.9.6 is going to be awesome!

mortonjt:
Welp.  Didn't even consider the taxa label color ...  Good catch!

While I was digging through the code I noticed this [line](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/emperor.js#L749).

Not sure what this function is for.  Regardless, I'm surprised that this doesn't error out, since currentColor isn't declared as a global variable

emperor/support_files/js/util.js
```diff
@@ -48,14 +48,13 @@ var k_CHROMABREWER_MAPNAMES = ['Classic QIIME Colors',
 /**
  *
  * Sorting function that deals with alpha and numeric elements
- * 
+ *
```
ElDeveloper: It seems like you modified the docstring for this function by accident, would you mind undoing the changes?

emperor/support_files/js/emperor.js
```diff
@@ -870,17 +878,33 @@ function toggleLabels() {
...
+
+        $('#taxalabels').css('display','block');
+        $("#taxaLevel" ).html(taxonomy[taxaLevel]);
+        for(var key in g_taxaPositions){
+                // get the coordinate of this taxa sphere
+            	var coords = toScreenXY(g_plotTaxa[key].position,g_sceneCamera,$('#main-plot'));
+
+            	// labels are identified by the key they have in g_taxaPositions
+            	labelshtml += "<label id=\""+key+"_taxalabel\" class=\"unselectable labels\" style=\"position:absolute; left:"+parseInt(coords['x'])+"px; top:"+parseInt(coords['y'])+"px;\">";
+                labelshtml += truncateLevel(g_taxaPositions[key]['lineage'], taxaLevel);
```
ElDeveloper: I think there's an extra space at the beginning of this line.

emperor/support_files/js/util.js
mortonjt: I think I know what happened ...

My editor is configured to automatically remove whitespace every time I save.  
I think there was whitespace before, and my editor deleted it once I saved it.

You sure we want the whitespace?

emperor/support_files/js/emperor.js
mortonjt: Not sure what's going on there.  Its actually aligned in my editor ...

emperor/support_files/js/emperor.js
```diff
@@ -48,7 +48,7 @@ var g_keyBuilt = false;
 var g_screenshotBind;
 var g_separator_left;
 var g_separator_history;
-
+var g_taxaLabelColor;
```
ElDeveloper: After our offline chat, this can now be removed! Thanks!

ElDeveloper:
This looks very good, @antgonza would you mind having a quick look?

emperor/support_files/js/emperor.js
```diff
@@ -739,6 +738,14 @@ function colorChangedForTaxaSpheres(color){
 	}
 }
 
+function colorChangedForTaxaLabels(color){
+        // get the taxonomic assignments and append '_taxalabel' to
+        // retrieve all the labels belonging to a sphere in the plot
+        for(var key in g_taxaPositions) {
```
antgonza: I'm reworking this file so it uses better ways to loop by using _ (underscorejs), could you change this to:
```
_.each(g_taxaPositions, function(taxaPositions, key){
  $('#'+key+"_taxalabel").css('color', color);
});
```

antgonza: Functionality works fine and I can make this changes. Gonna erase this comments and merge this PR.

mortonjt:
Thanks!

ElDeveloper:
Thanks, this is an awesome addition.

PR #363, opened 2015-03-10
fixed typo

gregcaporaso:


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_363/make_emperor/

Jorge-C:
For some reason travis couldn't find its config file: `Could not find .travis.yml, using standard configuration.` and then build fails trying to use some ruby stuff.

gregcaporaso:
@Jorge-C, I think it's b/c this is a commit to gh-pages. I got screwed up by this last week on another repo thinking that it would just pull from master, but it appears that it doesn't.

PR #367, opened 2015-03-18
WIP: Improving performance

antgonza:
As discussed on today's meeting these are the initial commits to improve Emperor performance.


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_367/make_emperor/

emperor/format.py
```diff
@@ -54,23 +57,23 @@ def format_pcoa_to_js(header, coords, eigvals, pct_var, custom_axes=[],
...
-    if number_of_axes>len(coords[0]):
+    if number_of_axes > len(coords[0]):
         number_of_axes = len(coords[0])
 
     # validating that all the axes are above 0.01%, this accounts for really
     # small variations explained in some axes that end up being not practical
     # as the GUI has some problems when presenting those values on screen
-    valid_pcoalabels = len([i for i in pct_var if i>0.01])
-    if number_of_axes>valid_pcoalabels:
+    valid_pcoalabels = len([i for i in pct_var if i > 0.01])
```
mortonjt: Should we replace this with `filter` ?

Jorge-C: I'm curious, what is the rationale? List comprehensions are very pythonic and readable, unless the filtering function happens to be eg a builtin.

mortonjt: I was referring to [this filter function](https://docs.python.org/2/library/functions.html#filter)

But it certainly not critical to make this change.

emperor/support_files/js/emperor.js
```diff
@@ -125,136 +125,137 @@ function toggleScaleCoordinates(element) {
...
-		// the color has to be formatted as an hex number for makeLine to work
-		currentColor = g_plotVectors[sample_id].material.color.getHex();
+    // updating the position of a vertex in a line is a really expensive
+    // operation, hence we just remove it from the group and create it again
+    g_elementsGroup.remove(vector);
 
-		// updating the position of a vertex in a line is a really expensive
-		// operation, hence we just remove it from the group and create it again
-		g_elementsGroup.remove(g_plotVectors[sample_id]);
+    for (var vertex in vector.geometry.vertices){
```
mortonjt: Dumb question - why cant we use `_.each` here?

mortonjt:
By the way as a heads up, I'm working on refactoring displaying sample spheres, taxa sphere and labels.  Tomorrow, I'm going to push into Antonio's fork - to minimize potential backtracking.

emperor/format.py
antgonza: Yes, agree with Jorge. Remember that we are focusing on JS but these changes in python are due to either changes needed for JS or flake8.

antgonza:
Great, thanks!

emperor/format.py
mortonjt: Sounds good!

wasade: iirc the use of `filter` and `map` are frowned upon by pep8

emperor/format.py
```diff
@@ -94,18 +97,20 @@ def format_pcoa_to_js(header, coords, eigvals, pct_var, custom_axes=[],
     # write the values for all the spheres
     js_pcoa_string += '\nvar g_spherePositions = new Array();\n'
     for point, coord in zip(header, coords):
-        all_coords = ', '.join(["'P%d': %f" % (i+1,coord[i]) for i in range(number_of_axes)])
-        js_pcoa_string += ("g_spherePositions['%s'] = { 'name': '%s', 'color': "
-            "0, 'x': %f, 'y': %f, 'z': %f, %s };\n" % (point, point, coord[0],
-            coord[1],coord[2], all_coords))
+        all_coords = ', '.join(["'P%d': %f" % (i+1, coord[i]) for i in
+                               range(number_of_axes)])
+        js_pcoa_string += ("g_spherePositions.push({ 'name': '%s', 'color': "
```
wasade: While the emphasis is on the js, this is operation is quadratic. Suggest doing the list/join pattern

emperor/support_files/js/emperor.js
```diff
@@ -125,136 +125,137 @@ function toggleScaleCoordinates(element) {
...
-			operation(g_plotEllipses[sample_id].position.y, g_fractionExplained[g_viewingAxes[1]]),
-			operation(g_plotEllipses[sample_id].position.z, g_fractionExplained[g_viewingAxes[2]]));
-
-		// scale the dimensions of the ellipse
-		g_plotEllipses[sample_id].scale.set(
-			operation(g_plotEllipses[sample_id].scale.x, g_fractionExplained[g_viewingAxes[0]]),
-			operation(g_plotEllipses[sample_id].scale.y, g_fractionExplained[g_viewingAxes[1]]),
-			operation(g_plotEllipses[sample_id].scale.z, g_fractionExplained[g_viewingAxes[2]]));
-	}
+  _.each(g_plotEllipses, function(ellipse){
```
wasade: this can be reduced further to minimize indexing. it is not using `_.each` but is directly amendable. i apologize if my js is off... still very not comfortable with it. 

btw, it looks like for loops perform [better](http://stackoverflow.com/questions/11887450/each-vs-for-loop-and-performance) than `_.each`, no?

```javascript
vx0 = g_fractionExplained[g_viewingAxes[0]];
vx1 = g_fractionExplained[g_viewingAxes[1]];
vx2 = g_fractionExplained[g_viewingAxes[2]];

for (sample_id in g_plotEllipses){
	pos = sample_id.position
	scale = sample_id.scale
	sample_id.position.set(operation(pos.x, vx0), operation(pos.y, vx1), operation(pos.z, vx2));
	sample_id.scale.set(operation(scale.x, vx0), operation(scale.y, vx1), operation(scale.z, vx2));
}
```

wasade: ...this type of strategy likely works in a few places, but am just putting the pattern up once. if people like this, i'll redo the rest of the loops. i don't think it'll be an incredible boost, but should cut down on overhead and compress the code a bit

ElDeveloper: I agree, this strategy will work in a few other places! Main benefit is
readability!!

mortonjt: :+1: if it is readable and gives a speed boost, then I'd think we'd definitely want this!

I think there are a few places that this will tricky to implement.  For example [drawTaxa](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/emperor.js#L1448).

From what I understand, we're treating associative arrays as dictionaries - meaning that the lookup time is linear.
So it would be nice to minimize [these sorts](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/emperor.js#L1458) of indexing operations

antgonza:
Just did another pass and we are getting really close!!

Things not working at this point (@mortonjt, could you take a look?):
- [ ] scaling
- [ ] labels dropdown menu 
- [ ] jackknifing opacity
- [ ] scale coords by percent explained 
- [ ] animations

mortonjt:
Awesome.  I'll look at it today

antgonza:
Thanks!

emperor/support_files/js/emperor.js
wasade: Great, thanks. Will issue commits against this PR today.

Just to reiterate though, I do not believe the use of `_.each` is a performance gain but instead can reduce performance. I think the _best_ performance we can get here is if we do classic C-style forloops, but that is from a very limited review of different iteration methods in js

ElDeveloper:
I'll take on scaling and animations (if that's ok with you guys).

antgonza:
Fine with me.

emperor/support_files/js/emperor.js
antgonza: _.each = @wasade is 100% right. This is my fault. At this point I
think we should go back to for loops, run tests here
https://jsperf.com/jquery-each-vs-each-vs-for-loop but keep the
improvements of the code within the loops.

mortonjt: :+1: But we should keep the Array of objects scheme right?

antgonza: Yup!

wasade: Starting on this now: the refactor of these for loops. This likely will cause conflicts

ElDeveloper:
I wanted to take this opportunity to mention that if possible, use two spaces to indent things in emperor.js, it currently is a horrible mess of spaces and tabs but if possible try to use two spaces.

For those of you using vim this should do the trick both for JS and HTML:

```viml
autocmd BufRead,BufNewFile *.js set tabstop=2                                   
autocmd BufRead,BufNewFile *.js set shiftwidth=2                                
autocmd BufRead,BufNewFile *.html set tabstop=2                                 
autocmd BufRead,BufNewFile *.html set shiftwidth=2    
```

mortonjt:
:+1: Not going to lie, the indentation has been a bit of a nuisance

wasade:
Ah, sorry! Will revert to two

ElDeveloper:
It's ok, since we don't actually have a standard for that file, it's
hard to strictly follow something.

wasade:
I propose we replace all ^I with 2 or 4 spaces, and clean up when we see
crap. I think that should be a separate PR as it'll get ugly, it impacts
like 1200 lines

ElDeveloper:
I agree on both respects doing it in a separate PR and just replacing ^I
with two spaces.

antgonza:
:+1:

emperor/support_files/js/emperor.js
```diff
@@ -985,17 +987,17 @@ function filterKey() {
...
 			$('#'+divid).css('display','none');
 		}
-	}
+  });
 }
 
 /*This function handles events from the ellipse opacity slider*/
-function ellipseOpacityChange(ui) {
-	document.getElementById('ellipseopacity').innerHTML = ui.value + "%";
+function ellipseOpeacityChange(ui) {
```
mortonjt: this change is incorrect :)

emperor/support_files/js/emperor.js
```diff
@@ -821,21 +812,25 @@ function labelMenuChanged() {
...
 
 /*This function is called when a label color is changed*/
 function labelColorChanged(value, color) {
-	g_categoryName = document.getElementById('labelcombo')[document.getElementById('labelcombo').selectedIndex].value;
-	value = value.replace('_','');
-
-	for(var i in g_plotIds){
-		var sid = g_plotIds[i];
-		var divid = sid.replace(/\./g,'');
-		if(g_mappingFileData[sid][g_mappingFileHeaders.indexOf(g_categoryName)] == value){
```
mortonjt: Looks like this line is the dealbreaker - we need the indexOf operation :(

mortonjt:
@antgonza , just also noticed that the taxa label display is broken - the labels don't adjust position when rotated.

mortonjt:
Looks like the following is broken
* Individual sample highlighting
* Color updating functionality
* Scaling taxa coordinates by percent explained
* Comparison plots
* Animations

ElDeveloper:
Also add:

- [ ] Plot labels.

I'm working on that last one right now, hopefully I'll have a PR
sometime soon.

mortonjt:
Awesome!

ElDeveloper:
Please, let's not fix spaces->tabs until we are done working on the file. Otherwise conflicts are **impossible** to go through.

antgonza:
OK but I thought I already did this ...

ElDeveloper:
No wonder my current work was destroyed in a sea of conflicts :(

ElDeveloper:
After talking with @wasade, @mortonjt, @antgonza and @josenavas, we've decided that for the time being this PR will be closed and will essentially be decomposed in smaller more gradual changes.

We will first change from tabs->2-spaces in emperor.js, and then start stubbing out the core structures that are going to be used in JS to make the codebase respect the MVC. Please get in touch if you would like to get involved with this effort.

PR #374, opened 2015-03-23
Fix Indentation

mortonjt:


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_374/make_emperor/

Issue #342, opened 2015-02-12
Modernize codebase


ElDeveloper:
~~Add coveralls.~~
Add flake8/pep8 checks.
Remove `__version__`,`__author__` ... etc strings where possible.
Add BSD license header to all files.
Remove shebang from library files.

I'm tagging this issue as an easy fix, because the changes should be very simple, however it might take a while to do them all.

ElDeveloper:
- [ ] @josenavas will work on sort.py and test_sort.py
- [ ] @mortonjt will work on biplots.py and test_biplots.py 
- [ ] @ElDeveloper will work on format.py and test_format.py
- [ ] @wasade filter.py and test_filter.py

-------

The following have been unclaimed yet:

- [ ] parse.py and test_parse.py
- [ ] test_util.py (util.py should already be done).
- [ ] setup.py

josenavas:
I will take test_util.py (I do not haver permissions on this repo to modify your comment @ElDeveloper )

ElDeveloper:
@josenavas, updated, thanks!

josenavas:
working on setup.py

josenavas:
And working on the remaining one... parse.py and test_parse.py

mortonjt:
@wasade , I can take care of pep8ing `filter.py` and `test_filter.py`

PR #376, opened 2015-03-23
Pep8 util

josenavas:
Another partial for #342


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_376/make_emperor/

ElDeveloper:
Looks good, but would be good to have another pair of eyes go through this, @mortonjt or @antgonza would any of you mind having a look?

PR #377, opened 2015-03-23
flake8-ing setup.py

josenavas:
Partial #342


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_377/make_emperor/

PR #385, opened 2015-03-29
BUG: Fix bug where large file downloads would fail

ElDeveloper:


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_385/make_emperor/

PR #393, opened 2015-05-08
Dummy model

josenavas:
Stubs out the code for the Models and Views of the MVC pattern.


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_393/make_emperor/

PR #397, opened 2015-05-09
Decomposition model

josenavas:
Adds the Decomposition Model object and its tests!


emperor-helper:
The test build for this pull request can be found here: http://emperor.colorado.edu/pull_397/make_emperor/

PR #406, opened 2015-06-22
Adds the initial work done at the BD2K hackathon

ElDeveloper:
This pull request includes all the code that was added during the BD2K
hackathon, most notably it includes the new JavaScript objects and tests that
form part of the new JavaScript API. A new examples folder has been created
where we can show examples of how the API can be used to visualize ordination
plots.


emperor-helper:
The test build for this pull request can be found here: http://emperor.microbio.me/pull_406/make_emperor/

ElDeveloper:
This should be ready for a review. @josenavas, @mortonjt, I've also started working on the Emperor controller (for a subsequent pull request), but I think it may be better to not expand the size of this one anymore. Plus I foresee a lot of discussion with the `EmperorController` :dizzy_face:

josenavas:
I think this is ok as it is right now. Do we want 2 devs or with one is enough to merge in here?

wasade:
Just throwing this out there... would it be useful to explore TypeScript
here? (not in this PR, just in general with EMPeror)

ElDeveloper:
@wasade I think it would be good to at least give it a shot, though
definitely not in this PR haha :P

Side note, I am not expanding on the documentation for EmperorController
because as it currently stands the object will change a fair amount and
I don't think it makes sense to document something that ephemeral.

@josenavas, I would be OK with one reviewer in this PR as most of it was
written by several of us. Sadly, we are also short on active developers,
so it may be a bit hard to enforce this.

ElDeveloper:
Thanks @josenavas!

PR #407, opened 2015-07-07
ENH: Implement metadata coloring using the new API

ElDeveloper:
This pull request adds the necessary bits to recolor by a metadata category.
The most important thing to note is the usage of SlickGrid to allow the
interface to seamlessly present thousands of unique categories (I even tried
with a million and there were no performance problems with the table, the GL
canvas just couldn't handle that many samples though).

Most of the work has been focused in controller.js and in color-editor.js, this
last one is a class that we needed to implement to allow for SlickGrid to edit
color values.

Notes
=====

- The code does seem to to scale (yay?).
- The API can be improved to make this even faster as we are repeating a few
  operations, and this is only noticeable when you have ~20K unique samples
  (though not too bad).
- We don't have code specific for the view controllers, currently everything is
  crammed inside controller.js (this should be a future PR).
- A lot of this code is very new, so please do not hesitate to ask any
  questions about it.

Testing
=======

In the examples folder, there's a `build.py` script that takes as an only
argument, the number of samples to display on screen.

```bash
cd examples
build.py 100

open examples/new-emperor.html
```

Once you run build.py you can see the example in new-emperor.html.


emperor-helper:
The test build for this pull request can be found here: http://emperor.microbio.me/pull_407/make_emperor/

antgonza:
Check the plots created by emperor-helper ...

ElDeveloper:
Oh :/ yeah that's OK, thanks for catching that, I'll fix it! Though note that those plots are not using the new API so they are not *as* relevant to this PR as the files under the examples directory.

Yoshiki Vázquez-Baeza

josenavas:
I've to say that the error is awesome! hahaha

@antgonza this PR is going to a different branch, so until everything is in place I don't think those tests are gonna pass, since we are "destroying" the code and starting from scratch.

ElDeveloper:
@antgonza, I just fixed those problems, thanks for bringing that up.

I just want to note that it's going to be hard to maintain backwards compatibility between code-bases, and we never committed to that, so it should be ok. Note also that this won't go directly into master so no need to worry if things are not working so as long as the intended task of te PR is completed.

On that same note, @josenavas, @antgonza, @mortonjt, has anyone reviewed the code, more importantly the demo associated with this change? I'm very excited by the fact that we can now scale to any number of samples and that the interface is no longer a barrier.

mortonjt:
Looking over it now!

mortonjt:
Ummm...  I think there are some problems with the demo.
Everything is being squashed out of the screen.

It happens with both Emperor-helper and when I test it on my laptop on one of my datasets

emperor/support_files/js/controller.js
```diff
@@ -1,37 +1,177 @@
...
+    this.sceneViews[0].resize(0, 0, 0.5 * plotWidth, 0.5 * this.height);
+    this.sceneViews[1].resize(0.5 * plotWidth, 0, 0.5 * plotWidth, 0.5 * this.height);
+    this.sceneViews[2].resize(0, 0.5 * this.height, plotWidth, 0.5 * this.height);
+  }
+  else if (this.sceneViews.length === 4) {
+    this.sceneViews[0].resize(0, 0, 0.5 * plotWidth, 0.5 * this.height);
+    this.sceneViews[1].resize(0.5 * plotWidth, 0, 0.5 * plotWidth, 0.5 * this.height);
+    this.sceneViews[2].resize(0, 0.5 * this.height, 0.5 * plotWidth, 0.5 * this.height);
+    this.sceneViews[3].resize(0.5 * plotWidth, 0.5 * this.height, 0.5 * plotWidth, 0.5 * this.height);
+  }
```
mortonjt: This looks like alotta constants.  Think we can replace 0.5 with idk, some variable named x?

mortonjt:
Alright, I wasn't completely up to date.  Ignore my previous comments.

Ran Emperor with 50,000 points.  Though it was a little sluggish and took ~5 min to load, it is still very useful.  This is really impressive!

I'll need to read up a bit on Slick-grid and the other commits that have been made.  So I'll be gradually making comments over the next few days.

emperor/support_files/js/controller.js
```diff
@@ -1,37 +1,177 @@
...
+    throw Error('More than four views are currently not supported');
+  }
+
+  this.renderer.setSize(plotWidth, this.height);
+
+  // resize the grid according to the size of the container, since we are
+  // inside the tabs we have to account for that lost space, hence the 0.9
+  var gridWidth = this.$plotMenu.width() * 0.9,
+      gridHeight = this.$plotMenu.height() * 0.9;
+  $('#myGrid').width(gridWidth);
```
mortonjt: Same here.  Can we replace 0.9 with a variable?

emperor/support_files/js/controller.js
```diff
@@ -1,37 +1,177 @@
...
-    new ScenePlotView3D(this.renderer, [new DecompositionView(dm)], div_id, 0.5 * this.width, 0.5 * this.height, 0.5 * this.width, 0.5 * this.height)];
+  // default decomposition view uses the full window
+  this.addView();
+
+  this.buildUI();
+};
+
+EmperorController.prototype.addView = function() {
+  if (this.sceneViews.length > 4) {
+    throw Error('Cannot add another scene plot view');
```
mortonjt: Would anyone be opposed to adding in an immutable global variable `maxScenePlotViews=4`?  

Just in case we may want to add more scene plot views in the future, it'll be easier to tune all of these constants in a single place, rather than digging through all of the source files.

mortonjt:
Made a few comments.

Noticed that there are quite a few hard coded constants - just enough to be worrisome in the future.  
May want to consider replace these constants with immutable global variables, not only for the sake of readability, but also in case we would like to change these constants in the future.

emperor/support_files/js/controller.js
```diff
@@ -1,37 +1,177 @@
...
+  $('#keys').append($select);
+  $('#keys').append("<div id='myGrid'></div>");
+  $('#myGrid').width(gridWidth);
+  $('#myGrid').height(gridHeight);
+
+  var grid;
+  var columns = [
+    {id: 'title', name: '', field: 'color', sortable: false,
+     maxWidth: 25, minWidth: 25, editor: ColorEditor,
+     formatter: ColorFormatter},
```
mortonjt: Stupid question: is there any reason why `25` is chosen for `maxWidth` and `minWidth`?

ElDeveloper: @mortonjt and I will meet tomorrow to discuss this issues and figure
something out.

emperor/support_files/js/color-editor.js
```diff
@@ -0,0 +1,133 @@
+/**
+ *
+ * @author Yoshiki Vazquez-Baeza
```
antgonza: Just out of curiosity, are we adding these headers to all scripts? Just wondering as we remove them from the python code ...

antgonza:
I think the code looks fine, once the comments are addressed

emperor/support_files/js/color-editor.js
ElDeveloper: We removed these from the Python files but we didn't from the JavaScript
files, not sure what to do though. Others have suggestions?

ElDeveloper:
Thanks, if others can have a look that would be fantastic!

ElDeveloper:
This now depends on the PR that @mortonjt opened here https://github.com/ElDeveloper/emperor/pull/1 

If possible I would like to get this merged/reviewed sometime soon, please let me know how can I facilitate this.

mortonjt:
I'll take a look at it tonight.  Want to try to get these merged tomorrow?

ElDeveloper:
That would be amazing!

emperor/support_files/css/emperor.css
```diff
@@ -102,7 +98,7 @@ div.emperor-separator:hover {
 div#emperor-menu {
   height: 100%;
   font-family: Helvetica, sans-serif;
-  width: 26%;
+  width: 27%;
```
wasade: gotta bring back the classic

ElDeveloper: Hahaha, this really is just accounting for the fact that I removed the
sliding bar that allows users to resize the plot space and the tab view
space.

ElDeveloper: Though note that this will be coming back.

examples/build.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
```
mortonjt: Think we can add some documentation to specify exactly what this build file is?  This is just a template right?

examples/template.html
```diff
@@ -0,0 +1,92 @@
...
+    <script type="text/javascript" src="../emperor/support_files/js/util.js"></script>
+    <script type="text/javascript" src="../emperor/support_files/js/sceneplotview3d.js"></script>
+    <script type="text/javascript" src="../emperor/support_files/js/controller.js"></script>
+    <script type="text/javascript" src="../emperor/support_files/js/color-editor.js"></script>
+
+  </head>
+  <body>
+    <script type="text/javascript">
+      // we need to remove this callback from the chosen library
+      function chosen_highlight_callback() {};
```
mortonjt: Can we remove this callback?

ElDeveloper: We could, but the thing is that to support a feature in chosen,
@antgonza had to modify the original source and if this callback is not
present, the console is flooded with error messages. So the answer is,
yes but not right now.

mortonjt: Alrighty, just wondering

examples/build.py
ElDeveloper: Right, this is just a quick way to generate random data that can be
visualized in Emperor, good for simple tests. I've added some
documentation, so please let me know if you would like me to expand on
something.

examples/template.html
ElDeveloper: :+1:

examples/build.py
mortonjt: :+1:

mortonjt:
I don't see any outstanding problems.  @antgonza , @josenavas, mind if you could take another look at this before merging?

mortonjt:
bump

emperor/support_files/js/color-editor.js
antgonza: I think removing everywhere is the way to go ...

antgonza: Should this be another issue?

ElDeveloper: I would suggest leaving this for another issue, lots of deletions,
perhaps worth it to leave as a future issue, as it is rather small and
mechanical.

Issue #143, opened 2013-06-28
Add procrustes vector coloring selector


ElDeveloper:
A new controller or tab should be added to change the coloring of the arrows used to connect samples when coordinates are being compared.

PR #413, opened 2015-08-01
ENH: Adds base work on a view controller - depends on #407

ElDeveloper:
Note that the approach that I took here is rather different than what we originally planned. Essentially what will result of this is class `AttributeViewController` that can take care of 'color', 'opacity' or 'visibility', since these are soooooo similar between each other, I took this approach, however I'm down to hear whatever you guys think would be better.

The main relevant file is `view-controller.js`.

~~Note, this PR depends on #407~~

### DIFF
--- a/emperor/support_files/js/view-controllers.js
+++ b/emperor/support_files/js/view-controllers.js
@@ -1,0 +1,163 @@
+/**
+ *
+ * @author Yoshiki Vazquez-Baeza
+ * @copyright Copyright 2013--, The Emperor Project
+ * @credits Yoshiki Vazquez-Baeza
+ * @license BSD
+ * @version 0.9.51-dev
+ * @maintainer Yoshiki Vazquez-Baeza
+ * @email josenavas@gmail.com
+ * @status Development
+ *
+ */
+
+/*
+ * @name AttributeViewController
+ *
+ * @param {Node} container, Container node to create the controller in.
+ * @param {String} attrName, Attribute name to control, can be one of 'color',
+ * 'opacity' or 'visibility'.
+ * @param {Float} width, Width in pixels for the controller;
+ * @param {Float} height, Height in pixels for the controller;
+ * @param {ScenePlotView} sceneViews, SceneViews to controll; Maybe remove and
+ * add a callback???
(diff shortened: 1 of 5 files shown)


emperor-helper:
The test build for this pull request can be found here: http://emperor.microbio.me/pull_413/make_emperor/

ElDeveloper:
This PR is open for bike-shedding! Note that the generic view controller hasn't yet been integrated with the rest of the system but, this is what I had in mind, instead of a hierarchy of rather redundant classes.

ElDeveloper:
I just updated this PR to integrate the generic view controller in the main application (thus the huge blob of deletions), please have a look and let me know if this makes sense. If we keep following this pattern, all remaining tabs in the app should become either unique classes or part of the `AttributeController` class. I would advise looking at the interface for the `AttributeController`, mainly the way the class is being initialized, it feels too tightly coupled with the rest of the objects, and there are a few methods missing, as usual run `./build.py XXX` and open new-emperor.html to test this out.

emperor-helper:
Anyone wants to take a pass, or should I start? :ghost:

antgonza:
:+1: are we doing single or multiple reviews?

ElDeveloper:
We are doing as many reviews as possible, but at least 1 :). In this
case @mortonjt, @josenavas and me discussed more in-depth about the
design, and we decided to change this, though I haven't actually done
that. There's a bit more of discussion that needs to happen before that.

I'll be better at updating PRs with offline discussions!

emperor/support_files/js/view-controllers.js
```diff
@@ -0,0 +1,210 @@
...
+                                minWidth: scope.SLICK_WIDTH,
+                                editor: ColorEditor,
+                                formatter: ColorFormatter};
+    }
+    else if (scope.attrName === 'opacity') {
+      // TODO
+    }
+    else if (scope.attrName === 'visibility') {
+      // TODO
+    }
```
mortonjt: This will have to change, since we opted for multiple subclasses instead of one giant superclass

ElDeveloper:
Closing this PR, in favor of a new one that @mortonjt and me will work on.

:church: PR #143.

PR #415, opened 2015-08-11
Improve error message for erroneous biplot_fp case

adamrp:
@wdwvt1 ran into this issue, and the error message was not clear enough to diagnose easily.

### DIFF
--- a/scripts/make_emperor.py
+++ b/scripts/make_emperor.py
@@ -783,6 +783,12 @@ def main():
 
     # write the biplot coords in the output file if a path is passed
     if biplot_fp and taxa_fp:
+        if biplot_fp.endswith('/') or isdir(biplot_fp):
+            option_parser.error("Do not specify a path to a new (path ending "
+                                "in a slash) or existing directory for "
+                                "biplot_fp. The output file will be a "
+                                "tab-delimited text file.")
+
         # make sure this file can be created
         try:
             fd = open(biplot_fp, 'w')


antgonza:
Thanks, should taxa_fp have the same conditionals?

adamrp:
Ideally, the code in cli for checking the `new_filepath` option type:
```python
def check_new_filepath(option, opt, value):
    return value
```
would be better, but that won't happen :laughing:

I theoretically could add the same checks for the 4 input and output files (-m, --biplot_fp, -s, taxa_fp), but I think that the cause for confusion for most people is about whether *outputs* are files/directories, and there is less confusion about the inputs. Opinions? @ElDeveloper @wdwvt1

mortonjt:
:+1: for placing the error message into a function.  It'll prevent a bit of copying/pasting

ElDeveloper:
I think we should avoid any work that gets into "patching" qcli, so I'm
fine with the solution in place. Is the error better now @wdwvt1?

wdwvt1:
looks good to me!

antgonza:
I think the solution is to remove qcli and use click, right? I think a semi-patch is worst than no-patch ...

ElDeveloper:
Although I agree with that, I don't think we are going to be able to
port everything to click and keep backwards compatibility before
releasing anything else in the 0.9.x series, which we should do sometime
soon (lots of bugfixes we haven't released). I am hoping that after
0.9.6 we will go onto 1.0 and aggressively modify the CLI, replacing
qcli with click however QIIME users of the 1.9.x series won't have
access to that, thus it makes sense for this patch to exist.

antgonza:
What about leaving the bug open and fix during click transition? I really don't like semi-patches ...

antgonza:
Sorry for multiple messages. Note that my vote is to fully fix ...

ElDeveloper:
But -s is the only file that is individually written out by
make_emperor.py right? Not sure what else would need to be changed.

antgonza:
Good point, I got confused with `if biplot_fp and taxa_fp:` ... then I guess is a complete patch?

ElDeveloper:
Ok, so we are all :+1: on this correct?

antgonza:
OK with me.

ElDeveloper:
Thanks @adamrp

PR #428, opened 2015-11-28
WIP: Baseclass

mortonjt:
Here is the tentative code for the abstract base classes - EmperorViewControllerABC and EmperorAttributeABC.

Now, I'm not exactly sure about how we want to proceed with implementing the ColorViewController.  Implementing the ColorViewController may change how we want to modify these abstract base classes.

### DIFF
--- a/tests/javascript_tests/test_view_controller.js
+++ b/tests/javascript_tests/test_view_controller.js
@@ -122,6 +122,7 @@ $(document).ready(function() {
                   ['PC.635', 'YATGCTGCCTCCCGTAGGAGT', 'Fast', '20071112']];
       decomp = new DecompositionModel(name, ids, coords, pct_var, md_headers,
                                       metadata);
+<<<<<<< HEAD
       var dv = new DecompositionView(decomp);
       sharedDecompositionViewDict['pcoa'] = dv;
 
@@ -141,6 +142,26 @@ $(document).ready(function() {
                                       metadata);
       dv = new DecompositionView(decomp);
       sharedDecompositionViewDict['biplot'] = dv;
+=======
+
+      // Slickgrid
+      var columns = [
+        {id: "pc1", name: "pc1", field: "pc1"},
+        {id: "pc2", name: "pc2", field: "pc2"},
+        {id: "pc3", name: "pc3", field: "pc3"},
+      ];
+
+      var options = {
+        enableCellNavigation: true,
+        enableColumnReorder: false
+      };
+      var data = [];
+      data.push({'pc1':1, 'pc2':1, 'pc3':1});
+      data.push({'pc1':1, 'pc2':1, 'pc3':2})
+
+      grid = new Slick.Grid("#id1", data, columns, options);
+
+>>>>>>> cc2e14edbd5d323ead2116b7e7916a00d7d28fa1
     },
     teardown: function(){
       sharedDecompositionViewDict = undefined;
@@ -153,8 +174,9 @@ $(document).ready(function() {
    *
    */
   test("Constructor tests", function(assert) {
-
+    var dv = new DecompositionView(decomp);
     var container = $('<div id="does-not-exist"></div>');
+<<<<<<< HEAD
     var controller = new EmperorAttributeABC(container, 'foo', 'bar',
                                              sharedDecompositionViewDict);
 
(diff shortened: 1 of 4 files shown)


tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
...
+
+      var options = {
+        enableCellNavigation: true,
+        enableColumnReorder: false
+      };
+      var data = [];
+      data.push({'pc1':1, 'pc2':1, 'pc3':1});
+      data.push({'pc1':1, 'pc2':1, 'pc3':2});
+
+      grid = new Slick.Grid("#id1", data, columns, options);
```
ElDeveloper: Where is the element `#id1` coming from?

tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
...
+  /**
+   *
+   * Test set metadata field
+   *
+   */
+  test('Test setMetadataField', function(){
+    var dv = new DecompositionView(decomp);
+    var container = $('<div id="does-not-exist"></div>');
+    var attr = new EmperorAttributeABC(container, 'foo', 'bar', 
+                                       {'scatter': dv}, 'butter');
```
ElDeveloper: We don't need butter here, right? See the following piece of text:

```javascript
/**
 *
 * @name EmperorViewControllerABC
 *
 * Initializes an abstract tab for attributes i.e. shape, color, size, etc.
 * This has to be contained in a DOM object and will use the full size of that
 * container.
 *
 * @params {Object} [decompViewDict] This is object is keyed by unique
 * identifiers and the values are DecompositionView objects referring to a set
 * of objects presented on screen. This dictionary will usually be shared by
 * all the tabs in the application. This argument is passed by reference.
 *
 * @return {EmperorAttributeABC} Returns an instance of the EmperorAttributeABC
 * class.
 *
 */
function EmperorAttributeABC(container, title, description, decompViewDict){
  EmperorViewControllerABC.call(this, container, title, description);

  if (decompViewDict === undefined){
// ....
```

tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
...
+   */
+  test('Test getSlickGridDataset', function(){
+    var dv = new DecompositionView(decomp);
+    var container = $('<div id="does-not-exist"></div>');
+    var attr = new EmperorAttributeABC(container, 'foo', 'bar', 
+                                       {'scatter': dv, 'biplot': dv}, 
+                                       'butter', grid);    
+    deepEqual(attr.getSlickGridDataset(), [{'pc1':1, 'pc2':1, 'pc3':1}, 
+                                           {'pc1':1, 'pc2':1, 'pc3':2}]);
+
```
ElDeveloper: Can you remove this line.

tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
...
+    var attr = new EmperorAttributeABC(container, 'foo', 'bar', 
+                                       {'scatter': dv, 'biplot': dv}, 
+                                       'butter', grid);    
+    deepEqual(attr.getSlickGridDataset(), [{'pc1':1, 'pc2':1, 'pc3':1}, 
+                                           {'pc1':1, 'pc2':1, 'pc3':2}]);
+    attr.setSlickGridDataset([{'pc1':1, 'pc2':2, 'pc3':3}, 
+                              {'pc1':1, 'pc2':1, 'pc3':2}])
+    deepEqual(attr.getSlickGridDataset(), [{'pc1':1, 'pc2':2, 'pc3':3}, 
+                                           {'pc1':1, 'pc2':1, 'pc3':2}]);
+
```
ElDeveloper: Can you remove this line.

tests/javascript_tests/test_view_controller.js
ElDeveloper: BTW, if you agree with this change, we have to remove this from all the other initializations.

tests/javascript_tests/test_view_controller.js
mortonjt: That's coming from one of the divs in `index.html`

tests/javascript_tests/test_view_controller.js
mortonjt: Probably not :/

tests/javascript_tests/test_view_controller.js
ElDeveloper: Hmmmm, I don't think it's good practice to create divs as part of
index.html, we should create them and destroy them as part of setup and
teardown.

mortonjt: Sounds good

tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
+$(document).ready(function() {
+
+  module("EmperorViewControllerABC", {
+
+    setup: function(){
```
ElDeveloper: since this setup and teardown are unused I think they should be removed.

tests/javascript_tests/test_view_controller.js
```diff
@@ -0,0 +1,330 @@
...
+
+    throws(function(){
+      controller.fromJSON('{foo:11}');
+    }, Error, 'Cannot call this abstract method');
+
+    throws(function(){
+      controller.toJSON();
+    }, Error, 'Cannot call this abstract method');
+  });
+
```
ElDeveloper: Can you remove this line?

ElDeveloper:
:+1: thanks @mortonjt

PR #431, opened 2015-12-27
ColorViewController [new-api]

ElDeveloper:
This PR introduces the `ColorViewController` class. The intent of this was not only to implement the controller but to also exercise and update the abstract base classes (`EmperorViewControllerABC` and `EmperorAttributeABC`).

This is a broad description of the changes:
- Updates to the base classes and tests.
- Moving coloring code from util.js to color-view-controller.js (the functions are now class methods and the constants are now class attributes).
- Creation and implementation of ColorViewController and tests.
- Update `EmperorController`: it now uses `ColorViewController`, has a helper method to create tabs (`addTabs`) and a few properties were renamed or added (`$plotSpaceId` -> `$plotSpace`, `_$tabsList`, and `_$tabContainer`).
- Updates template.html and index.html to use the new classes.
- Correct resizing of the content of the body section of the tabs (I think this looks really nice now).
- Updates `ScenePlotView3D` to use a decomposition view dictionary instead of an array, I've also added tests/docs that were missing in this class.

@mortonjt and I pair-programmed a fair bit of this, it would be **very** useful to have other eyes look over this.

Finally, there's a fair amount of "javascript voodoo" going on, so let me know if something is not exactly clear.


emperor/support_files/js/color-view-controller.js
```diff
@@ -0,0 +1,215 @@
...
+  EmperorAttributeABC.call(this, container, title, helpmenu,
+                           decompViewDict, options);
+  this.$header.append(this.$colormapSelect);
+
+  // the chosen select can only be set when the document is ready
+  $(function() {
+    scope.$colormapSelect.chosen({width: "100%", search_contains: true});
+    scope.$colormapSelect.chosen().change(options.categorySelectionCallback);
+  });
+
```
mortonjt: Extra space ?

emperor/support_files/js/controller.js
```diff
@@ -156,86 +194,48 @@ EmperorController.prototype.render = function() {
...
-    });
-  }(this));
-};
+  // set the identifier of the div to the one defined by the object
+  $('#' + id).attr('id', obj.identifier);
 
+  // now add the list element linking to the container div with the proper
+  // title
+  this._$tabsList.append("<li><a  href='#" + obj.identifier +
+                                       "-tab'>" + obj.title + "</a></li>");
```
mortonjt: over indent?

mortonjt:
Very minor style comments.

@antgonza , @josenavas would you two like to take a look at this?

emperor/support_files/js/color-view-controller.js
```diff
@@ -0,0 +1,214 @@
...
+  var SLICK_WIDTH = 25, name, value, colorItem;
+  var scope = this;
+
+  // this class uses a colormap selector, so populate it before calling super
+  // because otherwise the categorySelectionCallback will be called before the
+  // data is populated
+  this.$colormapSelect = $("<select class='emperor-tab-drop-down'>");
+
+  for (var i = 0; i < ColorViewController.Colormaps.length; i++){
+
```
antgonza: Remove?

emperor/support_files/js/color-view-controller.js
```diff
@@ -0,0 +1,214 @@
...
+ * @param {values} list of objects to generate a color for, usually a category
+ * in a given metadata column.
+ * @param {map} name of the color map to use, see ColorViewController.Colormaps
+ *
+ *
+ * This function will generate a list of coloring values depending on the
+ * coloring scheme that the system is currently using (discrete or continuous).
+*/
+ColorViewController.getColorList = function(values, map) {
+  var colors = {}, numColors = values.length-1, counter=0, interpolator,
```
antgonza: Not sure if we want to have in separate lines those vars that have an initial value and those that don't.

emperor/support_files/js/color-view-controller.js
```diff
@@ -0,0 +1,214 @@
...
+ * all the tabs in the application. This argument is passed by reference.
+ *
+ **/
+function ColorViewController(container, decompViewDict){
+  var helpmenu = 'Change the colors of the attributes on the plot, such as ' +
+                 'spheres, vectors and ellipsoids.';
+  var title = 'Color';
+
+  // Constant for width in slick-grid
+  var SLICK_WIDTH = 25, name, value, colorItem;
```
antgonza: A line for those vars with default values and other one for non initialized.

emperor/support_files/js/controller.js
```diff
@@ -156,86 +194,48 @@ EmperorController.prototype.render = function() {
...
+ *
+ * Helper method to resize the plots.
+ *
+ * @param {Array} [dvdict] Dictionary of DecompositionViews.
+ * @param {function} [viewConstructor] Constructor of the view controller.
+ *
+ **/
+EmperorController.prototype.addTab = function(dvdict, viewConstructor){
+  // nothing but a temporary id
+  var id = "" + Math.round(1000000 * Math.random());
```
antgonza: Can you add a comment about this being faster than #.tostring() or other ways and perhaps a link to something that holds that?

antgonza:
This looks great. Some minor comments. I think it's fine to merge and then fine tune once we have actual plots.

mortonjt:
We do have actual plots - the ColorViewController is completely functional.

@ElDeveloper, do you think it would be worthwhile to update emperor-helper?  
Or should we do that later?

antgonza:
Where?

mortonjt:
Take a look at examples/index.html

antgonza:
Thanks. Then :+1: after the comments are addressed.

emperor/support_files/js/controller.js
ElDeveloper: That's a good point, we don't have any specific motivation to do it that
way, so I think it makes more sense to make it readable, I've changed
this to be `toString`.

ElDeveloper:
Awesome, thanks for the review, I think I've addressed the changes, note
also that build.py will still generate a tiny plot with as many spheres
as you want. i.e. "./examples/build.py 100" will create a plot with 100
random samples.

Issue #421, opened 2015-09-02
options to seletively hide animations of some trajectories


RNAer:
In cases of many trajectories, it can be distracting to look at them all together. It would be awesome to have little checkboxes on each trajectories on the right panel to hide/show them.

![screen shot 2015-09-02 at 10 59 27 am](https://cloud.githubusercontent.com/assets/1000056/9639602/b5a9f642-5161-11e5-87a5-27cc60722b2c.png)

ElDeveloper:
Sounds like a great idea, I've added this to the description of #405, so we can take care of it during the refactoring that's ongoing at the moment.

PR #432, opened 2015-12-29
implemented new form of animation where samples are only shown at a particular point in time without being traced by lines

ConstantinoSchillebeeckx:
Fix for https://github.com/biocore/emperor/issues/421

### DIFF
--- a/emperor/_format_strings.py
+++ b/emperor/_format_strings.py
@@ -307,7 +307,7 @@ document.getElementById("logotable").style.display = 'none';
                 </tr>
                 <tr>
                     <td>
-                        <input type="checkbox" id="animate_traces" checked><label title="When checked, animated samples will trace lines on the plot as they move; otherwise samples will be plotted without a trace." >Render animation with traces.</label></input>
+                        <input type="checkbox" id="animate_traces" onchange="colorAnimationsByCategoryChanged()" checked><label title="When checked, animated samples will trace lines on the plot as they move; otherwise samples will be plotted without a trace and will be colored as defined in the 'Colors' tab." >Render animation with traces</label></input>
                     </td>
                 </tr>
                 <tr>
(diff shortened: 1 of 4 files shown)


ElDeveloper:
Thanks for submitting this @ConstantinoSchillebeeckx! It seems like the tests are failing, did these work in your system? You can check this by running `./tests/all_tests.py`, if you want to run the javascript tests, you will need to have `phantomjs` installed.

Additionally, it seems like there's some tests missing for the js code being added. I also noticed there's a mix of tabs and spaces in the JS code, so I would try to homogenize this and get it to be the same as the surrounding code (which should be 2 spaces for all the JS, but there may be some exceptions).

Let me know if you need any help.

ConstantinoSchillebeeckx:
This is my first time submitting some sort of pull-request and wasn't aware of the tests to be run.  I've run them on my end and some do, indeed fail.  I'm afraid I'm unsure why they fail as I'm unsure how they work.  I'm going to need some guidance on this ...

ElDeveloper:
The tests being executed are all found under the `tests` folder in the base directory. So have a look at those, specifically, I think the culprits are `test_format.py`. and `test_core.py` (those are related to your changes AFAIK). For more information have a look at the [contributing guidelines](https://github.com/biocore/emperor/blob/master/CONTRIBUTING.md).

On the JavaScript side of things, it seems like there's a few things failing, you can either open `tests/javascript_tests/index.html` in your browser (there are two tests that fail in some browsers, these are found in the `test_ui.js` but you shouldn't worry about that as those pass just fine with phantom.js in our CI server), this should help you figure out what other things are failing.

Finally, it seems that you have changed a lot of tabs and spaces in code that you are not modifying, I would suggest to revert this change, and only limit the modifications to the changes you are proposing. This will help reviewing and maintaining the code.

ConstantinoSchillebeeckx:
Just wanted to drop a note that these unit tests, formatting and git branching is all above my head - I'm therefore giving up on it.

For those interested, a saved version of the modified animation format can be found [here](https://drive.google.com/file/d/0B4Ou4lclm6GQcHRFZXFFSmsyekk/view?usp=sharing).

PR #435, opened 2016-01-07
Updates to css

mortonjt:
Depends on #434 

Also having a minor issue with the horizontal scrollbar.  It seems to work once the console is opened up.

Todo
- [x] Fix the css coloring
- [x] Fix the horizontal scrollbar issue
- [x] Fix the vertical scrollbar issue
- [ ] Fix elements disappearing from the grid (happens upon resize on the tabs that are not on-focus).
- [ ] Fix double clicking scheme
- [ ] Fix missing tick.png in ipython notebook

PR #434, opened 2016-01-06
Visibility controller

antgonza:
~~These are the basic changes to the color controller so we can add the visibility controller. I'm creating a WIP PR so @mortonjt can start his part from this version.~~

PR #436, opened 2016-01-07
Use require.js in the project

ElDeveloper:
We are now fully require.js compliant, I've updated the library code, the tests and the files under examples to use require.js. Consequently I've also added require.js to the project and to the licenses file.

Note that there are a lot of files being changed, but they all have very similar changes, essentially adding a couple lines where we specify the dependencies we want to import in that module, and then a line specifying the namespace these dependencies will use.

I had to do a fair amount of updates on the test suite, most notably, I had to introduce a couple of `asyncTest` as they were stochastically failing in the browser. I also noticed that we were misusing the `setup` and `teardown` methods, so I made a few changes to avoid using global variables.

Finally note that the indentation for almost all files that have been **require**-ified is a bit off i.e. most of the code should be indented a couple of spaces, but I decided that this is an easy thing to update once we have this first PR merged. Otherwise the diff would be too obnoxious to review.


ElDeveloper:
I need to update from #434 before this can be merged.

ElDeveloper:
This should be ready for another review. A **very rough** demo of Emperor embedded in the notebook can be found [here](http://nbviewer.ipython.org/gist/ElDeveloper/57fe8feb11b46dc58add). Note that we still have some CSS leaks, and an image problem (checkmarks of the visibility tab won't appear on screen). I'll fix that in an upcoming PR i.e. when we introduce the proper Python API.

wasade:
Should it be visible? I see the notebook but not the emperors... Although,
I'm on my phone right now

antgonza:
I just tried on laptop and phone and they work fine. Not sure why doesn't show in @wasade's phone. Code looks fine. I think it will be OK to merge this and then fix all css in https://github.com/biocore/emperor/pull/435. @mortonjt, what do you think?

wasade:
Probably internet connection... on ghetto roaming with my phone

wasade:
:+1: on merge. @eldeveloper, do you want to create an issue about the
missing images?

antgonza:
I'm gonna merge, added the image issue in https://github.com/biocore/emperor/pull/435 list.

ElDeveloper:
Awesome, thanks a bunch guys! @wasade, I think the reason why that may
have happened, is because of require.js' async nature.

PR #437, opened 2016-01-09
Fix require.js indentation

ElDeveloper:
After migrating to require.js, I left the indentation unchanged to make the review-ability of #436 more bearable. This PR is adding the missing two spaces to all the libraries and fixing a few docstrings to fit under 80 columns.

--------

I think reviewing using GitHub's "hide whitespace changes" feature, might be useful in this case, see [here](https://github.com/blog/967-github-secrets).


antgonza:
Merging as it's just indentation.

PR #439, opened 2016-01-17
FIX: Background

mortonjt:
#435 is turning into a nightmare to manage.

So this is a smaller PR that only handles the background color.  Suggest merging this before attempting other css changes.


antgonza:
Just to confirm, the solution was to add some missing figures and update the libraries ?

ElDeveloper:
I don't think we need bootstrap.css. Also if we now have a "selected" theme, do we still need the rest of the images in emperor/support_files/vendor/css/images like ui-bg_gloss-wave_35_f6a828_500x100.png or ui-icons_ef8c08_256x240.png or ui-icons_ffd27a_256x240.png ? I don't exactly know how the themes work and all but I think we may just not need these images at all.

mortonjt:
:+1: on removing those files.  It doesn't look like we need them anymore.

Yup.  That's exactly what is going on @antgonza

PR #442, opened 2016-01-24
Scrollbars

mortonjt:
Fixes the scrollbar issues in #435

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -162,19 +162,25 @@ define([
 
     this.renderer.setSize(plotWidth, this.height);
 
+    // resize the grid according to the size of the container, since we are
+    // inside the tabs we have to account for that lost space, hence the
+    var tabWidth = this.$plotMenu.width() * this.GRID_SCALE,
+    tabHeight = this.$plotMenu.height() * this.GRID_SCALE;
+
+    // the tab list at the top takes up a variable amount of space and
+    // without this, the table displayed below will have an odd scrolling
+    // behaviour
+    tabHeight -= this._$tabsList.height();
+    console.log('True width: ' + tabWidth);
+
     if (this.colorController !== undefined){
-      // resize the grid according to the size of the container, since we are
-      // inside the tabs we have to account for that lost space, hence the
-      var tabWidth = this.$plotMenu.width() * this.GRID_SCALE,
-      tabHeight = this.$plotMenu.height() * this.GRID_SCALE;
-
-      // the tab list at the top takes up a variable amount of space and
-      // without this, the table displayed below will have an odd scrolling
-      // behaviour
-      tabHeight -= this._$tabsList.height();
-
       this.colorController.resize(tabWidth, tabHeight);
     }
+
+    if (this.visibilityController !== undefined){
+      this.visibilityController.resize(tabWidth, tabHeight);
+    }
+
   };
 
   /**
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -201,6 +201,8 @@ define([
    *
    **/
   EmperorController.prototype.buildUI = function() {
+    var scope = this;
+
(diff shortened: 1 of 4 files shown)


examples/template.html
```diff
@@ -124,6 +124,10 @@
             animate();
           });
 
+          $(document).ready(function(){
```
ElDeveloper: How is the callback on line 113 different from this?

emperor/support_files/js/controller.js
```diff
@@ -162,19 +162,24 @@ define([
 
     this.renderer.setSize(plotWidth, this.height);
 
-    if (this.colorController !== undefined){
-      // resize the grid according to the size of the container, since we are
-      // inside the tabs we have to account for that lost space, hence the
-      var tabWidth = this.$plotMenu.width() * this.GRID_SCALE,
-      tabHeight = this.$plotMenu.height() * this.GRID_SCALE;
+    // resize the grid according to the size of the container, since we are
+    // inside the tabs we have to account for that lost space, hence the
```
ElDeveloper: I just noticed this comment is a bit off, though it was like this before, would you mind deleting the "hence the" at the end of this line?

ElDeveloper:
Thanks so much @mortonjt, though I'm seeing an odd scrolling in the visibility tab, see the animated gif below:

tab-scroll

Notice how I can go past the metadata selectors are scrolled up, aside from that, I think this is looking good.

I remember we had fixed this in your previous branch, right?

examples/template.html
mortonjt: This is something that I've been experimenting with.

`$(document).ready(function()){}` seems to be called when all of the visible elements are loaded.
This actually loads after `$(function()){}` is called.

I am not sure why this is the case - it seems counterintuitive, especially considering that they are supposed to be equivalent (at least according to SO).

I encourage you to test this out.

mortonjt:
No.  We never fixed that problem.

examples/template.html
ElDeveloper: Hmmm, should we get rid of all occurrences of `$(function())...` in favor of `document`? at least that way we would guarantee that we only display stuff once we truly are ready to display it.

ElDeveloper:
I seem to recall the problem with the visibility tab was that we somehow could only scroll through n-3 elements of the total n.

mortonjt:
I think @ElDeveloper and I just fixed this problem.  @josenavas @antgonza can you have a look and maybe merge?

emperor/support_files/js/controller.js
antgonza: This still needs to be addressed, right?

antgonza:
1 comment.

emperor/support_files/js/controller.js
ElDeveloper: yes!

josenavas:
Looks good once @ElDeveloper 's comments is addressed.

examples/template.html
mortonjt: Just tested that out.  If we move everything into document.ready(), it breaks.
```javascript
          $(document).ready(function(){
             init();
             animate();
             ec.resize(window.innerWidth, window.innerHeight);
          });
```
Here's something funny - if I have two declarations of document.ready(), it works again
```javascript
          $(document).ready(function(){
             init();
             animate();
          });

          $(document).ready(function(){
             ec.resize(window.innerWidth, window.innerHeight);
          });
```

mortonjt:
Comments addressed

examples/template.html
ElDeveloper: Can we just remove this? Or is that not possible?

mortonjt: Not possible.

mortonjt:
Who wants to merge this :)

josenavas:
Are there :beers: as a prize?

mortonjt:
Blazing fast! +1 :beer: for @josenavas

PR #444, opened 2016-02-04
Add axis lines to ScenePlotView3D

ElDeveloper:
This PR adds the axis lines to the `ScenePlotView3D` object, along with the necessary tests and updates to the decomposition model object.

The strategy here is to get the ranges over each dimension in every decomposition model object, then get all the decomposition model objects in a scene plot view, and union their ranges to get the final range.

-------

I had initially planned to submit a PR for the AxesViewController class, but I think this PR is already big as it is and we might get lost in the details. Also I'm not adding the percentage explained to the plot yet. :fork_and_knife:

### DIFF
--- a/emperor/support_files/js/model.js
+++ b/emperor/support_files/js/model.js
@@ -298,7 +298,7 @@ function ($, _) {
    *
    * Helper function used to find the minimum and maximum values every
    * dimension in the plottable objects. This function is used with
-   * uderscore.js' reduce function (_.reduce).
+   * underscore.js' reduce function (_.reduce).
    *
    * @param {accumulator} Object with a "min" and "max" arrays that store the
    * minimum and maximum values over all the plottables.
--- a/emperor/support_files/js/model.js
+++ b/emperor/support_files/js/model.js
@@ -86,6 +86,17 @@ function ($, _) {
    *
    * @class Represents all the information that need to be plotted
    *
+   * @property {String} [abbreviatedName] abbreviated name of the decomposition
+   * object being visualized.
+   * @property {Array} [ids] array of strings with the sample identifiers.
+   * @property {Array} [percExpl] array of floats with the percent explained by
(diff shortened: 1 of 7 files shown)


mortonjt:
Awesome!! Reviewing now.

emperor/support_files/js/model.js
```diff
@@ -276,6 +296,35 @@ function ($, _) {
...
+   * @param {plottable} Plottable object to compare with.
+   *
+   * @return Updated version of accumulator, integrating the ranges of the
+   * newly seen plottable object.
+   *
+   **/
+  DecompositionModel._minMaxReduce = function(accumulator, plottable){
+
+    // iterate over every dimension
+    _.each(plottable.coordinates, function(value, index){
```
antgonza: _.min(numbers); ?

antgonza:
One comment.

emperor/support_files/js/model.js
```diff
@@ -276,6 +296,35 @@ function ($, _) {
 
   /**
    *
+   * Helper function used to find the minimum and maximum values every
+   * dimension in the plottable objects. This function is used with
+   * uderscore.js' reduce function (_.reduce).
```
mortonjt: uderscore.js -> underscore.js

emperor/support_files/js/model.js
ElDeveloper: I don't think we can use `min` here, because we don't have access to vectorized operations over the whole matrix of coordinates, that's why we iterate over each one of them. I realize this is a bit confusing, but `_.reduce` is every plottable and for every plottable it finds the minimum and maximum over each axis, that's why we use `_.each` here.

mortonjt: :+1: I think it is fine as is.

mortonjt:
:+1:

PR #446, opened 2016-02-11
Add percentage explained labels

ElDeveloper:
This PR adds the necessary methods to add and remove the axes labels.

I noticed we could abstract out a few patterns so I made a couple helper methods for the add and remove methods. I also noticed we had a problem in the template.html file with a missing dependency declaration.

### DIFF
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -168,9 +168,9 @@ define([
       [range.min[x], range.min[y], range.max[z]]
     ];
 
-    for (var i = 0; i < 3; i++){
-      action(start, ends[i], i);
-    }
+    action(start, ends[0], 0);
+    action(start, ends[1], 1);
+    action(start, ends[2], 2);
   };
 
   /**
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -4,6 +4,7 @@ define([
     "draw"
 ], function (THREE, OrbitControls, draw) {
   var makeLine = draw.makeLine;
+  var makeLabel = draw.makeLabel;
   /**
    *
    * @name ScenePlotView3D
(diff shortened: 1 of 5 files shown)


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -162,13 +168,86 @@ define([
       [range.min[x], range.min[y], range.max[z]]
     ];
 
+    for (var i = 0; i < 3; i++){
```
antgonza: If you are always doing 3 calls, will it be easier to simply add the 3 lines here?

antgonza:
One comment, it looks good. Note that the size of the text doesn't resize with the size of the spheres. In other words, if you plot 5 the text is "readable" but if you plot 1000 is not.

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: I prefer the loop, but I am willing to change it if you guys thinks that makes more sense.

ElDeveloper:
Thanks for the review. I've also noticed that, and I think the way to
fix this will be by determining the sphere size according to the density
of samples per cubic measurement. From this sphere size we should then
determine the size of the text, does that make sense? If you recall from
our previous implementation, we were using some heuristics in the
Python-side of things, but I think we now have the structure and
information to be able to automatically do this in JavaScript. What do
you guys think?

emperor/support_files/js/sceneplotview3d.js
antgonza: Not strong feelings but I think we are using both in different places. However, it seems like if we are doing something with x, y, z we are being explicit and if we are doing 1, 2, 3 we are using for loops.

antgonza:
Automatically in JS sounds perfect

ElDeveloper:
Got it, I've addressed your comment, about the loop. As for the sphere sizes/label sizes, would you be ok with leaving that for a different PR?

antgonza:
Fine with me. Still doing 2 reviews? @mortonjt

ElDeveloper:
Great, and yes, would be awesome if @mortonjt could have a look.

mortonjt:
Looking awesome!

ElDeveloper:
thanks!

PR #448, opened 2016-02-18
BUG: Remove misleading QIIME version

ElDeveloper:


antgonza:
:+1: will merge when tests pass

PR #451, opened 2016-03-21
ENH: Revamping Python API

mortonjt:
Not sure how to handle the `BASE_URL` - getting a 404 there
Also getting the following error
```
VM114:3 Uncaught ReferenceError: $ is not defined(anonymous function) @ VM114:3(anonymous function) @ main.min.js:20168x.event.dispatch @ main.min.js:88y.handle @ main.min.js:88(anonymous function) @ main.min.js:88x.fn.extend.domManip @ main.min.js:88x.fn.extend.append @ main.min.js:88OutputArea._safe_append @ main.min.js:20210OutputArea.append_display_data @ main.min.js:20331OutputArea.append_output @ main.min.js:20116OutputArea.handle_output @ main.min.js:20030iopub.output @ main.min.js:21673Kernel._handle_output_message @ main.min.js:23263x.isFunction.i @ main.min.js:87Kernel._handle_iopub_message @ main.min.js:23290Kernel._finish_ws_message @ main.min.js:23127(anonymous function) @ main.min.js:23118
VM115:2 Uncaught ReferenceError: requirejs is not defined
```
- [ ] Unittests

### DIFF
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -23,15 +23,13 @@ Classes
 # ----------------------------------------------------------------------------
 from __future__ import division
 
-from emperor.format import (format_mapping_file_to_js, format_pcoa_to_js,
-                            format_taxa_to_js, format_vectors_to_js,
-                            format_comparison_bars_to_js,
-                            format_emperor_html_footer_string)
-from emperor._format_strings import EMPEROR_HEADER_HTML_STRING
-
+from template import TEMPLATE_HTML, TEMPLATE_JS
+from jinja2 import Template
+import numpy as np
 # we are going to use this remote location to load external resources
-RESOURCES_URL = 'http://emperor.microbio.me/master/make_emperor/emperor_outpu\
-t/emperor_required_resources'
+# RESOURCES_URL = 'http://emperor.microbio.me/master/make_emperor/emperor_outpu\
+#t/emperor_required_resources'
+#BASE_URL = 'https://cdn.rawgit.com/ElDeveloper/emperor/require-js'
 
 
 class Emperor(object):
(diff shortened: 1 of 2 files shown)


emperor/template.py
```diff
@@ -0,0 +1,260 @@
...
+
+  var dm, ec;
+
+  function init() {
+    // Initialize the DecompositionModel
+    dm = new DecompositionModel(name, ids, coords, pct_var,
+                                md_headers, metadata);
+    // Initialize the EmperorController
+    ec = new EmperorController(dm, 'python-penguin');
+    lol = ec
```
ElDeveloper: :see_no_evil:

emperor/template.py
```diff
@@ -0,0 +1,260 @@
+TEMPLATE_HTML = """
```
ElDeveloper: Since these are templates, instead of having them as static strings, we can have them as files. Would you mind saving these as separate files i.e. one for `TEMPLATE_HTML` and one for `TEMPLATE_JS`, then put them in the support_files directory, and finally make sure that they are listed in the setup.py file?

emperor/core.py
```diff
@@ -109,35 +105,30 @@ def _repr_html_(self):
...
-        # build the HTML string
-        output = [EMPEROR_HEADER_HTML_STRING, mf_string, pcoa_string, taxa,
-                  bars, vectors, footer]
-
-        # add the remote resources
-        _emperor = '\n'.join(output)
-
-        self._html = _emperor
+        # template = Template(TEMPLATE_STRING)
+        def listify(a):
```
ElDeveloper: I made a few changes to the code in `examples/build.py` recently, it fixes a few minor annoyances with the object translation, I would suggest that you copy and paste from there.

emperor/template.py
mortonjt: :+1: I'll remove it in the next commit, once I get a prototype working.

ElDeveloper: Great, thanks!

emperor/template.py
wasade: :rainbow: :cat2:

ElDeveloper:
Just FYI everyone, @mortonjt and me met IRL and figured out a fix for the problem we were observing with the URL :fireworks:

PR #452, opened 2016-03-23
Python API (based on the new-api branch)

ElDeveloper:
This PR introduces the starting point for the new Python API, main difference is that we now take as input a `DataFrame` object with the metadata, and a scikit-bio `OrdinationResults` object with the matrix data. The result is an object compatible with the rich-display infrastructure that the Jupyter notebook provides, have a look at a working example here (use chrome):

http://nbviewer.jupyter.org/gist/ElDeveloper/b218ad5b869c2cc91e8e

The notebook reflects the docstring in the object's class.


mortonjt:
Thanks for putting this together.

A couple of things that need to be discussed

1. Don't we want to make Emperor Python3 compatible.  Right now, it is not - it fails with `from string import strip`.  This will become a huge blocker when it comes into incorporating newer scikit-bio features in the future.
2. This api is reliant on an outdated version of scikit-bio.  (i.e. the OrdinationResult result is imported from scikit-bio=0.2.3).

ElDeveloper:
I agree, these are valid points:

1.- Let me see how hard it would be to make this PY3K compatiable, main
problem would be to update _everything_ else.

2.- Yes outdated indeed! I would like to explore the possibility of
supporting both versions, what do you guys think?

mortonjt:
Since this is for a demo, we can raise the above two points as issues.

Now, the doc-test isn't working.

1. `mf` and `ores` are backwards
2.  There is no index in `mf`
3.  I'm getting a `require.js?v=6da8be3…:900 ReferenceError: None is not defined(…)`

Below is the code that I tried testing.  Are you able to replicate this error?

```python
from core import Emperor
from skbio.stats.ordination import OrdinationResults
import pandas as pd
import numpy as np
data = [['Control', '20061218', 'Control_mouse_I.D._354'],
['Control', '20061218', 'Control_mouse_I.D._355'],
['Control', '20061126', 'Control_mouse_I.D._356'],
['Control', '20070314', 'Control_mouse_I.D._481'],
['Control', '20071210', 'Control_mouse_I.D._593'],
['Fast', '20071112', 'Fasting_mouse_I.D._607'],
['Fast', '20080116', 'Fasting_mouse_I.D._634'],
['Fast', '20080116', 'Fasting_mouse_I.D._635'],
['Fast', '20080116', 'Fasting_mouse_I.D._636']]
columns = ['Treatment', 'DOB', 'Description']
index = ['PC.354', 'PC.355', 'PC.356', 'PC.481', 
         'PC.593', 'PC.607', 'PC.634', 'PC.635', 
         'PC.636']
mf = pd.DataFrame(columns=columns, data=data, index=index)
    
eigvals = np.array([0.47941212, 0.29201496, 0.24744925,
                    0.20149607, 0.18007613, 0.14780677,
                    0.13579593, 0.1122597, 0.])
n = eigvals.shape[0]
site = np.random.randn(n, n)
site_ids = ('PC.636', 'PC.635', 'PC.356', 'PC.481', 'PC.354', 'PC.593',
            'PC.355', 'PC.607', 'PC.634')
p_explained = np.array([0.26688705, 0.1625637, 0.13775413, 0.11217216,
                        0.10024775, 0.08228351, 0.07559712, 0.06249458,
                        0.])
ores = OrdinationResults(eigvals, site=site, site_ids=site_ids,
                         proportion_explained=p_explained)
Emperor(ores, mf)
```

ElDeveloper:
Yes, thanks for catching that! This should work now, I'm surprised about
the require.js problem, can you see if this fixes it?

mortonjt:
screen shot 2016-03-23 at 5 33 39 pm

ElDeveloper:
@mortonjt, that code has an unnamed index in the pandas dataframe, if you change it to be the example in the docstring, then it should work just fine.

ElDeveloper:
@mortonjt, I just pushed a fix to change that.

ElDeveloper:
@mortonjt and by that I mean, try the same code without changing it, it should now account for cases with unnamed index vectors in the mapping file.

mortonjt:
Beautiful

emperor/core.py
```diff
@@ -128,9 +128,16 @@ class Emperor(object):
     def __init__(self, ordination, mapping_file, dimensions=5):
         self.ordination = ordination
 
+        self.mf = mapping_file.copy()
+
+        # the index of the dataframe needs to be named
+        if self.mf.index.name is None:
+            self.mf.index.name = 'SampleID'
```
mortonjt: :-1: Because this will still thrown an error if the index.name doesn't correspond to 'SampleID'.

I think it is better to just remove this line and throw and reasonable error.

mortonjt: I misread this.  I think is ok.

ElDeveloper:
Don't understand why some builds are failing :confused:, let me see if I can figure it out.

mortonjt:
I'd be ok with ditching set_index.  That looks like where the problem is originating.

ElDeveloper:
Great, tests pass now ... turns out it was an olde version of pandas with a bug that would not keep renamed index vectors, we were mainly seeing this because we are using conda and test against older numpy versions that limit the versions of pandas that you can use. Anyways, it should work now. :white_check_mark:

ElDeveloper:
Thanks @mortonjt!

PR #454, opened 2016-04-10
Improve usability of Emperor in the notebook

ElDeveloper:
This pull requests changes a few minor things that make for a better interaction in the notebook. Please make sure you try this out locally in a jupyter notebook, this example should help you get started:

http://nbviewer.jupyter.org/gist/ElDeveloper/a77a327a7da77583b80725041803e057

This is a summary of the changes:
- Allow users to set remote/local resource loading (`remote` argument).
- Creation of `nbinstall`.
- Removes all unused styles from emperor.css
- **Removes the base_Url attribute from the require.js config in main-template.html**: we needed to do this because otherwise two applications using require would have conflicts loading resources.

## Open questions

- Should we warn users that when they use `remote=False` they won't be able to see the plot in nbviewer/github?
- Should we automatically run nbinstall? Seems rather aggressive.


ElDeveloper:
Anyone available to review this PR?

antgonza:
Tests are failing ... should we wait until they pass?

ElDeveloper:
The build is failing due to coveralls ... odd, I guess I have to improve the tests. Unit tests are passing though.

mortonjt:
The basic functionality is working.  But Category Name is truncated and the right tab is cut off ...

antgonza:
I think the code looks good but I'm not sure how we will add this to Qiita. In other words, how easy will be to replace the current Emperor with this new structure.

ElDeveloper:
Thanks for the reviews, answers to the comments here:

>  But Category Name is truncated ...

The `Category Name` thing can be fixed, I think my suggestion is to remove it altogether, what do you guys think?

> and the right tab is cut off ...

Yes, this is a problem, though it's not really cut off, it's improperly resized and you can scroll around it, if you can't scroll let me know. The real issue here is that we need to execute the resize callbacks once the whole application is ready, if we can do this, we'll be good, but I haven't found a way to do this.

> In other words, how easy will be to replace the current Emperor with this new structure.

Do you mean the calls to `make_emperor.py`? If so there's two alternatives:

1.- Once we have a CLI interface, we would just update the command call.
2.- We could use Emperor at the JavaScript level and embed the code directly in the source of Qiita. And we would just format the coordinates and mapping file from Qiita in a custom way.

mortonjt:
I'm ok with removing the Category Name. One option is to make the `Category Name` an option within the menu, sort of like the `Please select option`.

Yeah, that resize issue has been haunting us for months now ...

Again, I don't think that we should address `make_emperor.py` in this PR.

ElDeveloper:
Will remove it then.

I'll see what I can do about the resize.

And I guess we'll defer the script interface for a later PR then.

antgonza:
There are real failures. 

Emperor: My concern is that we have both uses (local: download_no-internet, and remote: view results) in Qiita. How is this going to work? I know we can always figure it out later but just want to make sure we keep it in mind.

ElDeveloper:
Got it, now I get the question, thanks for bringing this up. The solution would be to set the `self.base_url` property to a local path, though this makes the `remote` parameter a bit strange, because you set it upon initialization, but you can now load resources in one of three ways:

1.- Locally from the Notebook server.
2.- Remotely from the GitHub repository.
3.- Locally from a copy of the `emperor_required_resources` folder.

Which maybe means that we have to change the `remote` parameter to be something like: `use_case` or `location` or `resource_dir`, or we can probably leave it as `remote`, that takes either a boolean or a string, if it is a boolean, `False` means it loads resources from the notebook server, `True` means it loads resources from the GitHub repository and if it is a string, it copies the resources to that folder and tries to load them from there, what do you guys think?

Tests should pass now, both coverage and build-wise.

mortonjt:
Ok.  Just tested on my system. 

Few questions
1. Did you mean to drop the `Category Name` text?
2. The PC axes labels scale during rotation.  Was this intentional?
3. Is this supposed to work without internet connection?  I couldn't get it working without internet connection. I agree with @antgonza on this one.

ElDeveloper:
I believe this should be ready for review, took me a while to figure it out but the resizing is now working as it should, turns out we had a bug in the way we were propagating the resizing callbacks. Looks so symmetrical and nice now:

screen shot 2016-04-18 at 9 04 50 pm

ElDeveloper:
BTW, @squirrelo, @mortonjt & @antgonza, I think the amount of code in here is rather large already so if that's OK with y'all, I'll defer the Python changes (remote `True/False/URL`) to future PR.

squirrelo:
Sounds good to me, just open an issue and link it here so we all remember.

ElDeveloper:
> Sounds good to me, just open an issue and link it here so we all remember.

Good idea, see #458

ElDeveloper:
Is anyone available for review?

mortonjt:
I'm getting the following error
```
http://localhost:8925/emperor/support_files/vendor/js/images/tick.png Failed to load resource: the server responded with a status of 404 (Not Found)
```
I don't see the tick.png uploaded under `support_files`.  Can this be included in this PR?

squirrelo:
I do not know what that file is, as that error has been there since before I started working on the project. Could you point me to what vendor file this is part of?

mortonjt:
Also `Category Names` is still funky.  Not sure if this is supposed to be addressed here.

Finally, the `remote` parameter doesn't seem to be working - I cannot load up Emperor in an IPython notebook when remote=False.

@antgonza , do you remember where `tick.png` originated from?  That should have been part of the visibility controller.

antgonza:
https://github.com/biocore/emperor/pull/435

ElDeveloper:
Exactly, tick.png is outside of the scope of this PR. @mortonjt have you tried running nbinstall? See the notebook linked above.

Yoshiki Vázquez-Baeza

mortonjt:
Ok, let's not worry about tick.png ( although I don't completely understand
this problem).

Yeah, I ran nbinstall just like suggested...

ElDeveloper:
Can you please paste the error you are getting or upload a screenshot?

Yoshiki Vázquez-Baeza

mortonjt:
Looks like that error has disappeared :L

I think this is good to go.
On Apr 19, 2016 4:41 PM, "Yoshiki Vázquez Baeza" <notifications@github.com>
wrote:

Can you please paste the error you are getting or upload a screenshot?

Yoshiki Vázquez-Baeza


> On Apr 19, 2016, at 4:23 PM, Jamie Morton <notifications@github.com>
wrote:
>
> Ok, let's not worry about tick.png ( although I don't completely
understand
> this problem).
>
> Yeah, I ran nbinstall just like suggested...
> On Apr 19, 2016 4:15 PM, "Yoshiki Vázquez Baeza" <notifications@github.com
>
> wrote:
>
> > Exactly, tick.png is outside of the scope of this PR. @mortonjt have you
> > tried running nbinstall? See the notebook linked above.
> >
> > Yoshiki Vázquez-Baeza
> >
> > > On Apr 19, 2016, at 4:09 PM, Antonio Gonzalez <
notifications@github.com>
> > wrote:
> > >
> > > #435
> > >
> > > —
> > > You are receiving this because you authored the thread.

> > > Reply to this email directly or view it on GitHub
> > >
> >
> > —
> > You are receiving this because you were mentioned.

> > Reply to this email directly or view it on GitHub
> > <https://github.com/biocore/emperor/pull/454#issuecomment-212167449>
> >
> —
> You are receiving this because you authored the thread.

> Reply to this email directly or view it on GitHub
>

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/biocore/emperor/pull/454#issuecomment-212172392>

ElDeveloper:
#sasquatchBug

Yoshiki Vázquez-Baeza

ElDeveloper:
Thanks for reviewing

Yoshiki Vázquez-Baeza

squirrelo:
👍  from me

ElDeveloper:
Thanks all!

PR #455, opened 2016-04-15
Viridis colormap

squirrelo:
the viridis colormap to chroma.js and makes it available to emperor users.

PR in to chroma.js to add this directly to the library as [104](https://github.com/gka/chroma.js/pull/104) so when that goes in we can use the regular library version.


ElDeveloper:
👍

PR #456, opened 2016-04-16
Gradient view and more colormaps

squirrelo:
~~Requires/supersedes #455 depending on how we want to do reviews.~~ Merged

This updates the colorview controller to allow gradient-based coloring instead of single-step coloring. It also adds in every available colormap in chroma into emperor.

This also fixes an issue found where this.$body was pointing to the slickgrid div itself, instead of a container div for the body with the slickgrid in it. Fixing this allowed the switch between colorbar and discrete category colors in the menu.

screen shot 2016-04-15 at 19 18 27

screen shot 2016-04-15 at 19 18 47


emperor/support_files/js/color-view-controller.js
```diff
@@ -56,62 +55,108 @@ define([
     var SLICK_WIDTH = 25, scope = this;
     var name, value, colorItem;
 
+    // Create scale div and checkbox for whether using scalar data or not
+    this.$scaleDiv = $('<div></div>')
```
ElDeveloper: Can you describe these properties in the documentation above? Also maybe some of these need to be private properties, I'll leave that to you, let me know if you would like to talk through that.

emperor/support_files/js/color-view-controller.js
```diff
@@ -56,62 +55,108 @@ define([
...
+    this.$colorScale = $("<svg width='90%' height='40'></svg>");
+    this.$scaleDiv.append(this.$colorScale);
+    this.$scaled = $("<input type='checkbox' id='scaled'>");
```
ElDeveloper: Unfortunately we cannot use the id tag, because in a Jupyter notebook we may have `n` different plots, so this would conflict between plots.

emperor/support_files/js/color-view-controller.js
```diff
@@ -56,62 +55,108 @@ define([
...
-      // the second array has the identifiers used to generate the colors
-      value = ColorViewController.Colormaps[i];
-      name = ColorViewController.ColormapNames[i];
-      colorItem = $('<option>').attr('value', value).text(name);
-
-      scope.$colormapSelect.append(colorItem);
+    for (var i = 0; i < ColorViewController.Colormaps.length; i++) {
+      var colormap = ColorViewController.Colormaps[i];
+      //Check if we are in a new optgroup
+      if (colormap.type != currType) {
```
ElDeveloper: Can you use `!==`?

emperor/support_files/js/color-view-controller.js
squirrelo: Yeah, I gave it an id but don't use it anywhere. I can remove that.

emperor/support_files/js/color-view-controller.js
squirrelo: yup

emperor/support_files/js/color-view-controller.js
```diff
@@ -128,48 +173,79 @@ define([
    * category in a given metadata column.
    * @param {map} name of the color map to use, see
    * ColorViewController.Colormaps
+   * @param {discrete} Whether to treat colormap requested as a discrete set
+   * of colors or use interpolation to create gradient of colors
+   * @param {scaled} Whether to use a scaled colormap or equidistant colors for
+   * each value. Defaiult false.
```
ElDeveloper: Defaiult -> Default is

emperor/support_files/js/color-view-controller.js
```diff
@@ -128,48 +173,79 @@ define([
...
+      else if (scaled === true) {
+        colors[values[index]] = interpolator(Number(values[index])).hex();
+      } else {
+        colors[values[index]] = interpolator(counter / numColors).hex();
         counter = counter + 1;
       }
     }
 
-    return colors;
+    //if scaled, generate the scale and add to div.
```
ElDeveloper: can you add a space before the `if`?

emperor/support_files/js/color-view-controller.js
```diff
@@ -128,48 +173,79 @@ define([
...
+    if (scaled) {
+      var min = _.min(numericValues);
+      var max = _.max(numericValues);
+      var mid = (min + max) / 2;
+      step = (max - min) / 100;
+      var stopColors = [];
+      for (var s = min; s <= max; s += step) {
```
ElDeveloper: I believe you can replace some of this with underscore's range, see here:
http://underscorejs.org/#range

emperor/support_files/js/color-view-controller.js
squirrelo: fixed

emperor/support_files/js/view-controller.js
```diff
@@ -83,7 +83,7 @@ define([
     this.$header.css('margin', '0 auto');
     this.$header.css('width', '100%');
 
-    this.$body = $('<div name="emperor-view-controller-body"></div>');
```
ElDeveloper: I believe this should still be `emperor-view-controller-body`, the gridDiv in turn should have a different name i.e. `emperor-grid-div` or something like that.

ElDeveloper:
@squirrelo **THIS IS AMAZING**.

Just a few notes: there's an error in the JavaScript console, though I haven't been able to trigger this again, do you have any idea as to why this may be happening:

```javascript
color-view-controller.js:204 Uncaught TypeError: Cannot read property '0' of undefined
```

Some notes, the `Continuous values` checkbox should be hidden if the color set is in the `Discrete` group, once you switch over to one of `Converging` or `Diverging` it should then be shown.

I belive the `Converging` group should be renamed as `Sequential`.

The `Continuous values` label should read something different, or have an explanation of what it does, maybe we can ask others in the lab what would they call this.

Does it make sense to have the colormap show regardless of what your scheme is i.e. whether the checkbox is marked or not?

The colorbar has a bit of unused space to the right, I think it would make sense to expand that to fit the width.

Finally, would you mind adding tests for the new properties that you added?

squirrelo:
Right now the continuous values checkbox is disabled if you are in a discrete colormap, but if really wanted I can make it disappear completely.

For me, it doesn't make sense to show the colormap regardless, because the entire coloring of the data is changed depending on if the slickgrid or colormap is shown. It would get confusing, since if the slickgrid is shown the colors of the points would not match up with the colormap.

I left the space on the right because as you shrink the page, that space gets eaten up. This was the largest I could get the bar and still have it show completely when you shrink the page as small as it goes.

I've added/updated tests for the new and changed properties, but if any are missing I can add them.

emperor/support_files/js/color-view-controller.js
wasade: Would it make sense to have a plot-specific prefix that's a randomly generated tag provided to all IDs? Not sure if this would work or if the syntax is sane but:

```javascript
function id_singleton_prefix() { 
    if(global_prefix === null) {
        global_prefix = something_random;
    return global_prefix;
}

...
this.$scaled = $("<input type='checkbox' id=id_singleton_prefix() + 'scaled'>");
```

emperor/support_files/js/color-view-controller.js
squirrelo: I could do 
```javascript
var range = _.range(min, max+1, step);
for(var pos in range) {
stopColors.push(interpolator(range[pos]).hex());
}
```
but this seems to be more complicated than the current setup.

emperor/support_files/js/color-view-controller.js
```diff
@@ -56,62 +55,108 @@ define([
...
+
+          if (scaled) {
+            scope.setSlickGridDataset({});
+            scope.$gridDiv.hide();
+            scope.$scaleDiv.show();
+            scope.$colorScale.show();
+            scope.$colorScale.html(colorInfo[1]);
+          } else {
+            scope.setSlickGridDataset(data);
+            scope.$scaleDiv.hide();
```
wasade: this and the next two lines appear replicated in the if block, do these need to be branch specific (e.g., due to execution order)?

ElDeveloper:
> Right now the continuous values checkbox is disabled if you are in a discrete colormap, but if really wanted I can make it disappear completely.

Making it disappear would be better, the checkbox serves no purpose except when the colormap is appropriate.

> For me, it doesn't make sense to show the colormap regardless, because the entire coloring of the data is changed depending on if the slickgrid or colormap is shown. It would get confusing, since if the slickgrid is shown the colors of the points would not match up with the colormap.

Yes I understand what you are saying. Anyone else have any comments on this?

> I left the space on the right because as you shrink the page, that space gets eaten up. This was the largest I could get the bar and still have it show completely when you shrink the page as small as it goes.

Noted, perhaps the only thing needed is to center the colormap, in the space it has?

-----------

Thanks for making these changes!

emperor/support_files/js/color-view-controller.js
squirrelo: As far as I can tell in the codebase (and @ElDeveloper correct me if I'm wrong) nothing is done through IDs directly, but instead using the objects attached to the controllers, so this is probably not needed.

emperor/support_files/js/color-view-controller.js
```diff
@@ -128,48 +173,79 @@ define([
    * category in a given metadata column.
    * @param {map} name of the color map to use, see
    * ColorViewController.Colormaps
+   * @param {discrete} Whether to treat colormap requested as a discrete set
```
wasade: is there a default?

emperor/support_files/js/color-view-controller.js
squirrelo: They are reversed, so this is hiding the scale and showing the slickgrid, while the if above does the opposite.

emperor/support_files/js/color-view-controller.js
ElDeveloper: Yeah, you are right Josh, for these properties, there's no need to have
unique identifiers. However, we have these for the
EmperorViewControllers in general, we have an `identifier` property and
it follows a scheme similar to what Daniel suggests.

emperor/support_files/js/color-view-controller.js
```diff
@@ -128,48 +173,79 @@ define([
...
       map = chroma.brewer[map];
       interpolator = chroma.bezier([map[0], map[3], map[4], map[5], map[8]]);
+      if (scaled === true) {
+        //assuming we have numeric since scaled called
+        var numericValues = _.map(values, Number);
+        min = _.min(numericValues);
+        max = _.max(numericValues);
+        interpolator = chroma.scale(map).domain([min, max]);
+      }
     }
```
wasade: I'm not seeing where `interpolator` get's set if `discrete === true`. From my read of the if/elifs starting on line 215, you _might_ expect the function call to be attempted in that case but I'm also not sure how JS implements if/elif and whether multiple branches can be taken.

emperor/support_files/js/color-view-controller.js
squirrelo: Nope, since it is a crucial piece of info depending on the colormap used.

emperor/support_files/js/color-view-controller.js
```diff
@@ -221,34 +297,53 @@ define([
...
-  'Purple-Blue-Green', 'Spectral', 'Red-Yellow-Green', 'Red-Blue',
-  'Pink-Yellow-Green', 'Pink-Red-Green', 'Red-Yellow-Blue', 'Brown-Blue-Green',
-  'Red-Grey', 'Purple-Orange', 'Viridis'];
-  ColorViewController._colorbrewerDiscrete = ["#8dd3c7", "#ffffb3", "#bebada",
-  "#fb8072", "#80b1d3", "#fdb462", "#b3de69", "#fccde5", "#d9d9d9", "#bc80bd",
-  "#ccebc5", "#ffed6f", /*first list ends here*/ "#a6cee3", "#1f78b4",
-  "#b2df8a", "#33a02c", "#fb9a99", "#e31a1c", "#fdbf6f", "#ff7f00", "#cab2d6",
-  "#6a3d9a", "#ffff99", "#b15928"];
+  ColorViewController.Colormaps = [
+    {id: 'discrete-coloring-qiime', name: 'Classic QIIME Colors', type: 'Discrete'},
```
wasade: minor, but would it be possible to place the `"Discrete"`, `"Converging"`, and `"Diverging"` strings into single variables? Reduces the potential for a copypasta error now or in the future in the event these strings change

emperor/support_files/js/color-view-controller.js
squirrelo: There is no interpolator for discrete values, since no interpolation is needed. Like 215 is indeed where the check is made, and javascript respects if/elif and only does one branch. Switch is the only thing that can fall through as far as I'm aware.

emperor/support_files/js/color-view-controller.js
squirrelo: Done

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/view-controller.js
squirrelo: good point, changed.

emperor/support_files/js/color-view-controller.js
squirrelo: I've moved them to headers of each section as objects with `type` property. However, this means there are two different types of objects in this list, so that will need to be taken into account my anything else using this list.

squirrelo:
Comments addressed

ElDeveloper:
👏

Issue #458, opened 2016-04-19
Change the remote parameter to accept URLs


ElDeveloper:
The `remote` parameter should also accept URLs in the `Emperor` constructor, see #454

PR #461, opened 2016-04-26
ENH: Change remote to accept a custom URL

ElDeveloper:
As discussed in #454 and in #458, remote now accepts a custom URL instead of
just a boolean.

### DIFF
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -60,12 +60,13 @@ class Emperor(object):
         identifiers in the `ordination` object.
     dimensions: int, optional
         Number of dimensions to keep from the ordination data, defaults to 5.
-    remote: bool, optional
-        Whether to load resources from a remote URL or to do it from the
-        nbextensions folder in your Jupyter installation. Note you will need to
-        import the `nbinstall` function from the util module and then call it
-        so the installation of the resources can take place. Defaults to load
-        resources from a remote location.
+    remote: bool or str, optional
+        This parameter can have one of the following three behaviors according
+        to the value: (1) `str` - load the resources from a user-specified
+        remote location, (2) `False` - load the resources from the nbextensions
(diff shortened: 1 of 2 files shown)


emperor/core.py
```diff
@@ -60,12 +60,13 @@ class Emperor(object):
         identifiers in the `ordination` object.
     dimensions: int, optional
         Number of dimensions to keep from the ordination data, defaults to 5.
-    remote: bool, optional
-        Whether to load resources from a remote URL or to do it from the
-        nbextensions folder in your Jupyter installation. Note you will need to
-        import the `nbinstall` function from the util module and then call it
-        so the installation of the resources can take place. Defaults to load
-        resources from a remote location.
+    remote: bool or str, optional
```
antgonza: What about the case that you only want to have a stand alone plot? Like we currently have.

antgonza:
1 question.

emperor/core.py
ElDeveloper: That's the cool thing about this change, we can just specify a relative path and it should work! 

Yoshiki Vázquez-Baeza

wasade: Why differentiate local/remote? `file://foo/bar`, `https://foo/bar`, and `http://foo/bar` are valid

wasade: oh i see, whether to stage local or pull from remote...

ElDeveloper: Yeah you got it, and this is the problem exactly, the parameter name is a bit confusing :-/

Yoshiki Vázquez-Baeza

squirrelo:
Looks good to me.

mortonjt:
Just tested without wifi, and it works.

👍

ElDeveloper:
If this looks good, can someone merge?

PR #462, opened 2016-04-27
BUG: Fix problem with icons not showing in the nb

ElDeveloper:
Replace the images used by SlickGrid to be jQuery icons.

Fixes a point in #435


ElDeveloper:
This won't show in the source code, but I replaced this:
```html
<img src='../emperor/support_files/vendor/js/images/tick.png'>
```

For this:
```html
<div class='ui-state-default ui-corner-all'>
     <span class='ui-icon ui-icon-check'>
</span></div>
```

Which results in this:

screen shot 2016-04-27 at 12 09 37 pm

As you can see we now have crosses and ticks for the visibility tab.

ElDeveloper:
Do not merge, I just found a problem!

ElDeveloper:
Nix that, I wasn't able to reproduce the problem.  `¯\_(ツ)_/¯`, just for the record, if someone sees a horizontal scroll bar in one of the view controllers, let me know, that's the problem I saw for a second, and then I wasn't able to see or reproduce again.

squirrelo:
👍

ElDeveloper:
If this looks good can someone merge this?

PR #463, opened 2016-04-27
BUG: Allow multiple independent plots in 1 nb

ElDeveloper:
Since we were using an identifier-based way to retrieve the object, we
couldn't have more than one Emperor plot in a jupyter notebook. I've
changed the references to no longer use the id attribute.


squirrelo:
👍

ElDeveloper:
🔁 anyone want  to merge? 🔁

PR #465, opened 2016-04-28
Shift-Rightclick menu

squirrelo:
Adds rightclick context menu to the canvas div, and adds ability to save current canvas as PNG as first activity on the context menu.

rightclickmenu


wasade:
Why are two right clicks necessary?

squirrelo:
Better gif added to illustrate this, but the first is a right click on the canvas and save as png to open a new wondow with the PNG in it. The second is right click save image on that png.

wasade:
Ah, I see, thanks

emperor/support_files/js/controller.js
```diff
@@ -253,10 +254,42 @@ define([
...
+   *
+   **/
+  EmperorController.prototype.screenshot = function(type) {
+    type = type || 'png'
+    // Render all scenes so it's rendered in same context as save
+    for (var i = 0; i < this.sceneViews.length; i++) {
+      this.sceneViews[i].render();
+    }
+    var c = this.renderer.domElement.toDataURL("image/" + type);
+    var w = window.open('about:blank','image from canvas');
```
ElDeveloper: Can you think of a workaround without having to open a new window? We've been discouraged from using direct references to window by the SAGE2 developers, is it possible at all to dump as a file blob, that would also save users 2-3 clicks and a new tab.

emperor/support_files/js/controller.js
```diff
@@ -253,10 +254,42 @@ define([
...
 
   /**
    *
+   * Save the current canvas view to a new window
+   *
+   * @param {string} [type] What type to save the file as. Default png.
+   *
+   **/
+  EmperorController.prototype.screenshot = function(type) {
+    type = type || 'png'
```
ElDeveloper: Can you add a semicolon at the end of this line?

emperor/support_files/js/controller.js
```diff
@@ -253,10 +254,42 @@ define([
                                  scope.resize(scope.$divId.width(),
                                               scope.$divId.height());
                                }});
+
+    // Set up the context menu
+    this.$contextMenu = $.contextMenu({
+      selector: '.emperor-plot-wrapper',
+      items: {
+        'saveImage': {
+          name: 'Save Image (png)',
```
ElDeveloper: png -> PNG

emperor/support_files/js/controller.js
```diff
@@ -1,14 +1,15 @@
...
     "jquery",
     "underscore",
+    "contextmenu",
     "three",
     "view",
     "scene3d",
     "colorviewcontroller",
     "visibilitycontroller",
-    "shapecontroller"
+    "shapecontroller",
```
ElDeveloper: Can you remove this trailing comma?

emperor/support_files/js/controller.js
```diff
@@ -1,14 +1,15 @@
 define([
     "jquery",
     "underscore",
+    "contextmenu",
```
ElDeveloper: Can you add the license of contextmenu to LICENSES.md?

emperor/support_files/js/controller.js
```diff
@@ -253,10 +254,42 @@ define([
                                  scope.resize(scope.$divId.width(),
                                               scope.$divId.height());
                                }});
+
+    // Set up the context menu
+    this.$contextMenu = $.contextMenu({
+      selector: '.emperor-plot-wrapper',
```
ElDeveloper: Is there any way we can have this be the `this.$plotSpace` element, and not all elements with this class?

ElDeveloper:
Code looks fine, just a few minor comments. Do you have any suggestions on what we can do regarding the conflicting right-click + drag events?

When I tested this out in the notebook, I couldn't get to see the menu on the plot, it instead would be appended (kinda) the lower left corner of the notebook:

screen shot 2016-04-29 at 12 08 53 am

You can try this notebook out, just run `jupyter notebook` in a local environment where you have emperor installed, and run all cells (remember to *clear all output*, *save*, *reload page* and *re-run all cells* in the notebook every time you make changes to the JS code):
https://gist.github.com/66c236fd0c0d6a92606be6b6788bc800

This is shaping up really well!!

emperor/support_files/js/controller.js
squirrelo: I think I can get it to dump a blob now that the render issue has been solved by adding another dependency. I'll look into it.

emperor/support_files/js/controller.js
squirrelo: done

emperor/support_files/js/controller.js
squirrelo: done

emperor/support_files/js/controller.js
squirrelo: done

emperor/support_files/js/controller.js
squirrelo: There isn't a licences.md file. The readme.md simply states it's MIT licenced.

emperor/support_files/js/controller.js
squirrelo: Unfortunately it pnly takes a jquery selector string, and not a full jquery element (https://swisnl.github.io/jQuery-contextMenu/docs.html). Is there any element ID that is unique to each instance?

squirrelo:
Not sure about the rightclick+drag event, but the menu isn't showing up because it needs the CSS as well.

emperor/support_files/js/controller.js
squirrelo: Now downloads directly, without any new windows.

ElDeveloper:
> Why are two right clicks necessary?

This is no longer the case 💥

ElDeveloper:
There's a conflict between right-clicking and plot repositioning, as it stands we can no longer move the plot around.

I looked for alternatives to this, but wasn't able to figure a workaround, so I've opened an issue at their repo https://github.com/swisnl/jQuery-contextMenu/issues/391 once this is solved, we should be good to merge this. The code and functionality is good except for this conflict.

-------------

An alternative I thought of was holding shift to deactivate the contextmenu callback and allow for right-clicking to work as it used to (reposition the plot), but not sure what others think about this?

josenavas:
Why is this functionality using a menu with right click rather than in the menu on the right?
Alternatively, shift+right click as @ElDeveloper is proposing is a common usage in OpenGL/WebGL tools (or at least in the ones that I've interacted - developed)

ElDeveloper:
We chose right-click because using the tab on the right (as in the
previous version) was a bit of overkill and the "lazy solution",
right-click makes this feel a lot more natural, the idea is to continue
building up and add an option to save SVG, etc from the right click
menu.

I would be OK with using shift, but want to make sure others know about
this, as this is a significant change.

josenavas:
thanks for the explanation @ElDeveloper 

in that case I would advocate for the shift solution. The natural behavior of emperor is to be able to visualize the plot from multiple points of view. The image exporting, although being an important feature, is secondary compared with the visualization.

ElDeveloper:
So your suggestion is that the right-click menu should only be shown
when shift is being pressed? Thanks for chiming in! @antgonza,
@mortonjt, any thoughts? If you just pull from this branch and run
./build.py 10 in the examples folder, you should be able to interact
with the `new-emperor.html` plot that includes all these changes

antgonza:
I think shift+right-click is a good way to go.

wasade:
Agree

squirrelo:
shift-rightclick is now in place for opening the menu.

ElDeveloper:
Superb @squirrelo, thanks!

Thanks for the input @wasade, @josenavas & @antgonza!!

Issue #327, opened 2015-01-13
Click on points to reveal label


lkursell:
Sometimes the cluster of points is too dense to distinguish an individual point's label when the labels are all turned on. Would be nice to be able to click on one point and have the label appear, perhaps at the bottom of the PCoA plot, like what happened in KiNG

antgonza:
Agree, have you tried playing with the transparency of the points?

lkursell:
That helps some, and with that + zooming the answer is eventually retrievable…though this is not pretty to look at:

wasade:
This would be awesome but just so you're aware, it is potentially
challenging to implement. Hit boxes are surprisingly tedious to resolve and
may require data structure changes to put in place.

lkursell:
No worries, and thanks for the insight, Antonio’s suggestion worked

ElDeveloper:
:+1: on that this issue should be resolved! When I get some time I'll
try having a look and see how problematic this may be to implement.

clozupone:
Hey, what is the current status of this issue? This is the one functionality that I really miss from KING. Particularly with a large dataset, it is nearly impossible to tell from labels what sample a point represents.

rob-knight:
+1

ElDeveloper:
@clozupone we are doing a major refactoring at the moment (see #405), I've added this feature request as part of the points that must be covered, more updates soon.

PR #467, opened 2016-05-03
Callback on click

squirrelo:
This adds the ability to attach callbacks when points are clicked. 

callbakcs2

callbakcs


squirrelo:
Start of #327

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -100,6 +103,26 @@ define([
     this.dimensionRanges = {'max': [], 'min': []};
     this.drawAxesWithColor(0xFFFFFF);
     this.drawAxesLabelsWithColor(0xFFFFFF);
+
+    this.raycaster = new THREE.Raycaster();
```
ElDeveloper: Looking at the API of the class, maybe we have to set the `near` and `far` attributes to match up those of the camera?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -100,6 +103,26 @@ define([
...
+    this.mouse = new THREE.Vector2();
```
ElDeveloper: Does this need to be a public attribute of the instance? Does it need to be an attribute at all?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -50,7 +50,10 @@ define([
    *
    **/
   ScenePlotView3D = function(renderer, decViews, container, xView, yView,
-                             width, height){
+                             width, height, callback){
+    var scope = this;
+    // Make setting a callback optional
+    this.callback = callback || function(name, item) {};
```
ElDeveloper: The problem with the callback being an attribute, is that then we can only have one subscriber to the callback. What I think we should do is model this as a pub/sub pattern. Let's chat about this, it may just be easier if we pair-program this bit.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -100,6 +103,26 @@ define([
     this.dimensionRanges = {'max': [], 'min': []};
     this.drawAxesWithColor(0xFFFFFF);
     this.drawAxesLabelsWithColor(0xFFFFFF);
+
+    this.raycaster = new THREE.Raycaster();
+    this.mouse = new THREE.Vector2();
+
+    // Add callback call when sample is clicked
+    $container.on('click', function(event) {
```
ElDeveloper: This is overall good, except that we can't easily test the raycasting/projection logic, do you think the bits inside this function could be abstracted so testing becomes more natural?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -100,6 +103,26 @@ define([
...
+      event.preventDefault();
+      var element = scope.renderer.domElement;
+      scope.mouse.x = ((event.clientX - element.offsetLeft) / element.width) * 2 - 1;
+      scope.mouse.y = -((event.clientY - element.offsetTop) / element.height) * 2 + 1;
+
+
```
ElDeveloper: Can you remove this line?

ElDeveloper:
Looks good overall! Just a few comments.

emperor/support_files/js/sceneplotview3d.js
squirrelo: This is taken care of by the line `scope.raycaster.setFromCamera(scope.mouse, scope.camera);`

emperor/support_files/js/sceneplotview3d.js
squirrelo: The suggestion from [the docs](http://threejs.org/docs/#Reference/Core/Raycaster) and [stackoverflow](http://stackoverflow.com/a/18553739) suggest just instantiating it once and reusing it.

emperor/support_files/js/sceneplotview3d.js
squirrelo: done

emperor/support_files/js/sceneplotview3d.js
squirrelo: done and test added.

ElDeveloper: Thanks!

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: Noted, thanks! Can you add this as a "private" property then?

squirrelo: done

squirrelo:
Just as a heads up, there is no way to trigger a click event and have it work with the SVGRenderer used in the tests.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -100,6 +102,90 @@ define([
...
+      //dont do anything if no subscribers
+      if (subscribers[eventType].length === 0) {
+        return;
+      }
+
+      var element = scope.renderer.domElement;
+      mouse.x = ((event.clientX - element.offsetLeft) / element.width) * 2 - 1;
+      mouse.y = -((event.clientY - element.offsetTop) / element.height) * 2 + 1;
+
+      raycaster.setFromCamera(mouse, scope.camera);
```
wasade: Ray casting will get expensive and potentially inaccurate for a drag box. What you really want to do is ask the opengl engine for what is present in your view, and resolve the "hit" object or objects from there by navigating the object hierarchy. Details are [here](http://www.glprogramming.com/red/chapter13.html) although I'm not immediately sure how to translate this to threejs

wasade: this is where things like quad and octtrees become very useful

wasade: Given the discussion [here], and from other posts from googling, it looks like the selection and feedback modes are not available in webgl. Not surprised as it is an embedded gl. But, from that same discussion, and tracing the suggestions, it is to actually use [raycasting](https://github.com/mrdoob/three.js/blob/master/examples/webgl_interactive_cubes.html#L140). very weird and surprising.

wasade: Last comment, I swear. Just found this, which is definitely relevant and discusses webgl, selection boxes and the issues with raycasting.

wasade: I lied. Also very relevant. Uses the marquee selection from the last comment, and also shows how to combine caching

https://github.com/cubicleDowns/webgl-code-samples/tree/master/canvas-interaction

squirrelo: From this, it seems like the selection will not catch a single or double click, as you need to get the entire objet in the selection space to select it. What we can do is for click events use raycasting, and for mouseDown/mouseUp use the marquee. this way we can merge this PR for click events, and then easily add on the marquee as a single new helper function for the two events. Sound good?

wasade: I think that's reasonable. Would it be possible to decompose this logic to simplify the `eventCallback`? What might be nice is something like:

```javascript
 var eventCallback = function(eventType, event) {
    event.preventDefault();
    // don`t do anything if no subscribers
    if (subscribers[eventType].length === 0) {
        return;
    }

    switch(eventType) {
        case 'click':
        case 'dblclick':
            eventMouseSelectPoint();
            break;
        default:
            // should never get here...
            break;
    }
```

It may make sense to do some decomposition on the rest of the event handling system as well given that multiple event types will need to handled. This may be a larger conversation though... but probably painful to do later.

btw, it looks like the events are developer defined given like 152. Can these be part of a controlled vocabulary? It would be useful to make them more descriptive as well, e.g., `MOUSE_LEFT_CLICK` and `MOUSE_LEFT_DOUBLE_CLICK`.

squirrelo: This is actually already decomposed, as eventType is only to grab the list of function calls that need to be run and run them, and the actual switch-type thing as you are suggesting is line 148+, where the event callback is hooked up to the actual events. 

The `on` and `off` function names, as well as the event controlled vocabulary, are taken from jquery to keep things consistent wiht that library, since it's such a standard for javascript.

ElDeveloper: @wasade, I believe I understand what you are referring to regarding the decomposition of the event callbacks, but at this point, we are getting away by calling the same handler irregardless of the originating event i.e. `click` or `dblclick`, this happens to be OK because both handlers expect the same argument and are intended to be called on the same event i.e. a mesh is clicked/double-clicked; this is why we can just call all the subscribers for any given event type.

ElDeveloper:
Do not merge this, I just submitted a PR against @squirrelo's branch adding tests, etc. Let's wait until that is in before merging. PR here:
https://github.com/squirrelo/emperor/pull/2

ElDeveloper:
@wasade @antgonza @josenavas @mortonjt can you guys have a look. @squirrelo and I worked together on this, so it would be good to get another pair of eyes overlook this changeset.

This PR adds two methods to ScenePlotView3D, the `on` and `off` methods, these methods follow a pub/sub pattern, currently we only have two events that you can subscribe to, the 'click' and 'dblclick' (stands for double click) events. These events are triggered when a mesh is clicked or double-clicked, all subscribers receive the name (sample id) and the mesh object that represents the sample.

Note that the tests were particularly tricky as we have to use `SVGRenderer` in the test suite (so the tests run in a headless browser), and SVGRender doesn't quite support ray-casting (according to @squirrelo's experiments).

ElDeveloper:
Friendly ping @wasade, @antgonza, @josenavas & @mortonjt. 🙏

mortonjt:
Is this supposed to work within an IPython notebook?

squirrelo:
It should, although how you register callbacks in the notebook hasn't been
worked out yet.

ElDeveloper:
Yes, absolutely, though note that you would have to subscribe a
callback, specifically, note that the changes are reflected at the
JavaScript level i.e. there are no noticeable changes in the UI.
Reviewing the code in this case is more useful than the interface.

mortonjt:
I think this code makes sense.  👍

tests/javascript_tests/test_sceneplotview3d.js
```diff
@@ -252,5 +261,153 @@ requirejs([
...
+     * Test the 'click' callback is resolved
+     *
+     */
+    test('Verifying click works', function(){
+      // for the test to pass, two assertions should be made
+      expect(2);
+
+      var renderer = new THREE.SVGRenderer({antialias: true});
+      var spv = new ScenePlotView3D(renderer, this.sharedDecompositionViewDict,
+          'fooligans', 0, 0, 20, 20);
```
antgonza: This should be indented more to the right, right?

antgonza:
One minor comment.

wasade:
Seems fine?

ElDeveloper:
@antgonza, comment should be addressed and tests are passing.

ElDeveloper:
Thanks @antgonza!

PR #468, opened 2016-05-05
Non numeric categories colormap

squirrelo:
This adds the ability to do colormaps for mixed string and numeric columns. Useful if you have a numeric column with a few string types for missing data, or for highlighting missing values in what should otherwise be a numeric column.

An error alert is given if there are 1 or 0 numeric values, otherwise it does this:

(this gif is outdated):
coloring


examples/build.py
```diff
@@ -43,10 +45,10 @@ def listify(a):
 coords = np.random.randn(N, 10).tolist()
 pct_var = (1/np.exp(np.arange(10))).tolist()
 
-md_headers = ['SampleID', 'DOB']
+md_headers = ['SampleID', 'DOB', 'Strings']
 metadata = []
 for _id in coords_ids:
-    metadata.append([_id, ''.join(sample(categories, 1))])
+    metadata.append([_id, ''.join(sample(categories, 1)), ''.join(choice(ascii_letters) for x in range(10))])
```
ElDeveloper: Mind keeping this at 80 chars?

emperor/support_files/js/color-view-controller.js
```diff
@@ -132,16 +132,26 @@ define([
           var data = decompViewDict.setCategory(attributes, scope.setPlottableAttributes, category);
 
           if (scaled) {
-            scope.setSlickGridDataset({});
-            scope.$gridDiv.hide();
+            //build plottables as all non-numerics
```
ElDeveloper: Mind adding a leading space in this and all the other comments?

emperor/support_files/js/color-view-controller.js
```diff
@@ -132,16 +132,26 @@ define([
...
+            var nonNumeric = uniqueVals.filter(function(v) { return isNaN(v)});
```
ElDeveloper: I think you should be able to filter passing just the isNaN function.

emperor/support_files/js/color-view-controller.js
ElDeveloper: I think this whole block about determining non-numeric values and their respective plottables belongs in a separate function that we should be able to test, the edge cases here sound like a nightmare to have untested.

emperor/support_files/js/color-view-controller.js
```diff
@@ -132,16 +132,26 @@ define([
...
+            var plotList = data.filter(function(x) {
+              return $.inArray(x.category, nonNumeric) !== -1;
+            });
+            var plottables = [];
+            for (var i = 0; i < plotList.length; i++) {
+              plottables = plottables.concat(plotList[i].plottables);
+            }
+            // Set SlickGrid for color of non-numeric values and show color bar for rest
+            scope.setSlickGridDataset([{category: 'Non-numeric values', value:'#64655d', plottables: plottables}]);
+            scope.$gridDiv.css('height', '30px');
```
ElDeveloper: In general, we don't really want to have a hardcoded height, this makes it hard to maintain a consistent size after resizing callbacks. Why can't we use the hide/show methods?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
    * continuous).
    */
   ColorViewController.getColorList = function(values, map, discrete, scaled) {
-    var colors = {}, numColors = values.length - 1, counter = 0;
-    var interpolator, min, max;
+    var scope = this;
```
ElDeveloper: This is a class method, how is it that you need `scope` here? Is this vestigial from the refactoring? Also just noticed that `scaled` is being redefined below i.e. ti is part of the function's arguments and then is re-declared with var, since you are making changes to this file, would you mind changing that?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+    if (values.length == 1 && discrete === false) {
       colors[values[0]] = chroma.brewer[map][0];
       return [colors, gradientSVG];
     }
 
-    if (discrete === false) {
+    //Call helper function to create the required colormap type
+    if (discrete) {
+      colors = ColorViewController.getDiscreteColors(values, map);
+    } else if (scaled) {
```
ElDeveloper: Would you mind putting all the else blocks in a new line?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+      var info = ColorViewController.getScaledColors(values, map);
+      colors = info[0];
+      gradientSVG = info[1];
+    } else {
+      colors = ColorViewController.getInterpolatedColors(values, map)
+    }
+    return [colors, gradientSVG];
```
ElDeveloper: Can you update the docstring?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+   *
+   * This function will generate a list of coloring values depending on the
+   * coloring scheme that the system is currently using (discrete or
+   * continuous).
+   *
+   * Defaults to use qiime discrete colors if there's no map passed in.
+   *
+   */
+  ColorViewController.getDiscreteColors = function(values, map) {
+    var map = map || 'discrete-coloring-qiime';
```
ElDeveloper: You are redefining map here.

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+
+  /**
+   *
+   * Retrieve a discrete color.
+   *
+   * @param {values} list of objects to generate a color for, usually a
+   * category in a given metadata column.
+   * @param {map} string, name of the discrete color map to use.
+   *
+   * This function will generate a list of coloring values depending on the
```
ElDeveloper: Just noticed this block is out of date, the map is an argument and does not depend on the "system".

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
-    if (discrete === false) {
+    //Call helper function to create the required colormap type
+    if (discrete) {
+      colors = ColorViewController.getDiscreteColors(values, map);
+    } else if (scaled) {
+      var info = ColorViewController.getScaledColors(values, map);
+      colors = info[0];
+      gradientSVG = info[1];
+    } else {
+      colors = ColorViewController.getInterpolatedColors(values, map)
```
ElDeveloper: Can you add a semicolon at the end of this line?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+   *
+   * This function will generate a list of coloring values depending on the
+   * coloring scheme that the system is currently using (discrete or
+   * continuous).
+   *
+   * Defaults to use viridis colormap if there's no map passed in.
+   *
+   */
+  ColorViewController.getScaledColors = function(values, map, nanColor) {
+    var map = map || 'viridis';
```
ElDeveloper: Can you remove the `var` in this and the next line?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+  /**
+   *
+   * Retrieve a scaled color set with scale bar.
+   *
+   * @param {values} list of objects to generate a color for, usually a
+   * category in a given metadata column.
+   * @param {map} string, name of the discrete color map to use.
+   *
+   * This function will generate a list of coloring values depending on the
+   * coloring scheme that the system is currently using (discrete or
```
ElDeveloper: This block is also out of date, can you remove it? Also, can you add a `@return` section?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+   */
+  ColorViewController.getScaledColors = function(values, map, nanColor) {
+    var map = map || 'viridis';
+    var nanColor = nanColor || '#64655d';
+    map = chroma.brewer[map];
+
+    // Get list of only numeric values, error if none
+    var numericValues = values.filter(function(v) { return !isNaN(v)});
+    if (numericValues.length < 2) {
+      alert('less than 2 numeric values found, can not create scale for category!');
```
ElDeveloper: Can we remove the `alert` calls from these functions? These shouldn't care about the logic of the UI, these should be merely helpers.

ElDeveloper: Just to be more specific, what would be helpful is if we had appropriate try/catch blocks on the consumers of these functions.

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+    //build the SVG showing the gradient of colors for values
+    var mid = (min + max) / 2;
+    var step = (max - min) / 100;
+    var stopColors = [];
+    for (var s = min; s <= max; s += step) {
+      stopColors.push(interpolator(s).hex());
     }
-
+    var gradientSVG = '<defs><linearGradient id="Gradient" x1="0" x2="1" y1="1" y2="1">';
+    for (pos in stopColors) {
```
ElDeveloper: Just noticed this, can you change this iteration to use the indices instead of the `in` keyword?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+   * Defaults to use viridis colormap if there's no map passed in.
+   *
+   */
+  ColorViewController.getScaledColors = function(values, map, nanColor) {
+    var map = map || 'viridis';
+    var nanColor = nanColor || '#64655d';
+    map = chroma.brewer[map];
+
+    // Get list of only numeric values, error if none
+    var numericValues = values.filter(function(v) { return !isNaN(v)});
```
ElDeveloper: I'm a little bit worried about this whole block, we are doing multiple passes over the same array of values doing slightly different things at each time, is there a way we could do a single pass with a helper function?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+    var map = map || 'viridis';
+    map = chroma.brewer[map];
+
+    var total = values.length;
+    var counter = 0;
+    interpolator = chroma.bezier([map[0], map[3], map[4], map[5], map[8]]);
+    var colors = {};
+    for(var i = 0; i < values.length; i++) {
+      colors[values[i]] = interpolator(counter / total).hex();
+      counter++;
```
ElDeveloper: How is `counter` different from `i`?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
+   *
+   * This function will generate a list of coloring values depending on the
+   * coloring scheme that the system is currently using (discrete or
+   * continuous).
+   *
+   * Defaults to use viridis colormap if there's no map passed in.
+   *
+   */
+  ColorViewController.getInterpolatedColors = function(values, map) {
+    var map = map || 'viridis';
```
ElDeveloper: Can you remove the `var`?

emperor/support_files/js/color-view-controller.js
```diff
@@ -193,71 +204,154 @@ define([
...
 
   /**
+   *
+   * Retrieve an interpolatd color set.
+   *
+   * @param {values} list of objects to generate a color for, usually a
+   * category in a given metadata column.
+   * @param {map} string, name of the discrete color map to use.
+   *
+   * This function will generate a list of coloring values depending on the
```
ElDeveloper: Can you remove this block and add the `@return` section?

tests/javascript_tests/test_color_view_controller.js
```diff
@@ -126,7 +126,7 @@ requirejs([
 
     test("Test ColorViewController.getColorList exceptions", function(){
       var five;
-      five = [0, 1, 2, 3, 4];
+      five = ['0', 'string1', 'string2', 'string3', 'string4'];
```
ElDeveloper: I'm skeptical that we are testing all the separate functions by just adding a new test case 😆 can you add the appropriate test cases?

ElDeveloper:
Overall functionality looks good, though there are a few problems I noticed:

1.- Grid gets hidden after switching between tabs: if you turn your coloring scheme into a continuous "coloring" and then switch to another tab, then come back, you are going to see that the grid is now clipped:

screen shot 2016-05-07 at 5 23 43 pm

The problem persists even after changing the "continuous" coloring:

screen shot 2016-05-07 at 5 26 08 pm

Also, in testing this feature, I noticed that the labels for the colormap overlap when the precision is high:

screen shot 2016-05-07 at 5 28 12 pm

emperor/support_files/js/color-view-controller.js
squirrelo: I can, but I wanted to keep this consistent between all the filter functions.

emperor/support_files/js/color-view-controller.js
squirrelo: I need to have a specific height for a single row of the slickgrid, so you can still change the color of the non-numeric values. I will dig into slickgrid to see if I can pull that height dynamically, unless you know the command.

emperor/support_files/js/color-view-controller.js
squirrelo: all vestigial of refactoring and moving from protected to class method.

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/color-view-controller.js
squirrelo: The two passes is to do the validation that we do indeed have enough values to do the scaling. Not sure it can be done without two passes.

emperor/support_files/js/color-view-controller.js
squirrelo: Done.

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/color-view-controller.js
squirrelo: done

emperor/support_files/js/color-view-controller.js
squirrelo: vestigial from refactor. Removed.

tests/javascript_tests/test_color_view_controller.js
squirrelo: No actual functionality was changed save that error throwing, so the tests made sure it still works the same. I can add helper function tests though.

emperor/support_files/js/color-view-controller.js
ElDeveloper: I see, but I don't think we should do that in this case.

emperor/support_files/js/color-view-controller.js
ElDeveloper: If I recall correctly, from their documentation, the grid assumes the
size of the containing div, so we just have to make this div the size of
the available space, which is what should already be happening AFAIK.

emperor/support_files/js/color-view-controller.js
ElDeveloper: One way to do it would be to do a single scan where we search and count
things, if we could also sort as we are doing this, that would be
exactly what we want. Maybe we can sketch something out.

tests/javascript_tests/test_color_view_controller.js
ElDeveloper: I see. I think that suggestion is good, could probably mean just
re-purposing some of the already existing tests, and making sure all
paths are covered.

squirrelo:
Comments addressed. Fixed the shrinking SlickGrid issue and turned the colorbar 90 degrees to allow longer values:

screen shot 2016-05-09 at 13 32 36

ElDeveloper:
Thanks for making these changes @squirrelo, I noticed a few things though. The sidebar is getting an odd size when you switch over to continuous values, there should not be a scrollbar, but there is, no idea why:

screen shot 2016-05-09 at 11 10 32 pm

Also I noticed the grid and bar are overlapping in some cases, the solution for this would be to use an HTML spacer or something like that:

screen shot 2016-05-09 at 11 11 17 pm

Finally, one minor thing that I only noticed when I looked at a real dataset 👀, the low small values of the color bar are at the top and the high values at the bottom, should have noticed before, but what we want is likely low values on the bottom, high values on top.

ElDeveloper:
@squirrelo, I just noticed something else. I tried saving the setting for a plot where I had checked the "Continuous Values" checkbox, and set a different color for the missing values, and when I tried to load the settings, I got two errors:

screen shot 2016-05-09 at 11 21 21 pm

screen shot 2016-05-09 at 11 21 23 pm

Let me know if you need help reproducing this problem!

squirrelo:
Thanks for catching those. Fixed now, and ready to go.

emperor/support_files/js/color-view-controller.js
```diff
@@ -277,42 +394,50 @@ define([
...
+      // Get the current SlickGrid data and update with the saved color
+      var data = this.bodyGrid.getData();
+      data[0].value = json.data['Non-numeric values'];
+      this.setPlottableAttributes(decompViewDict, json.data['Non-numeric values'], data[0].plottables);
+
+    }
+    else {
+      var data = decompViewDict.setCategory(json.data, this.setPlottableAttributes, json.category);
+    }
+    console.log(data);
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/color-view-controller.js
```diff
@@ -277,42 +394,50 @@ define([
...
+    this.setSlickGridDataset(data);
   }
 
-  /**
-   *
-   * Retrieve a discrete color.
+    /**
```
ElDeveloper: Can you move this line back to how it was before?

ElDeveloper:
@squirrelo, the grid gets resized oddly. To reproduce, scale continuous values, then switch tabs, switch back turn off the scaling and then turning it on:

screen shot 2016-05-10 at 3 05 18 pm

Note that we can't just hide the scrollbars, we need to actually fix this so the size of the table is correct. Also, why is the colorbar using so much vertical space?

squirrelo:
Should be fixed now.

ElDeveloper:
@squirrelo, this is working great, works flawlessly, the only odd thing is the display of the grid and the colorbar:

screen shot 2016-05-12 at 9 54 08 pm

Note how the grid is clipped, this is more evident when changing colors, not sure what's going on, some light debugging on Chrome seems to hint at the fact that the grid is inferring a different size (not 100% sure), no idea what might be going on.

Also why is there a space between the colorbar and the non-numeric grid? With those two things in, I think this is 👍 💯

squirrelo:
I can't recreate the grid clipping, so unsure how to fix. The SVG is not obeying the height percentage given when in the notebook, but work as a standard index file. Will try to figure that out.

ElDeveloper:
When I saw this problem, I was running emperor in the Jupyter notebook
and switching between tabs.

Thanks!

Yoshiki.

squirrelo:
Issues should be addressed now.

emperor/support_files/js/color-view-controller.js
```diff
@@ -187,77 +219,162 @@ define([
    * @param {scaled} Whether to use a scaled colormap or equidistant colors for
    * each value. Default is false.
    *
+   * @return {colors} The object containing the hex colors keyed to each sample
```
ElDeveloper: I don't think we are allowed to have two `@return` tags, are we?

squirrelo: We are, it's `@returns` thats bad.

emperor/support_files/js/color-view-controller.js
```diff
@@ -187,77 +219,162 @@ define([
...
-        stopColors.push(interpolator(s).hex());
+  /**
+   *
+   * Retrieve a scaled color set with scale bar.
+   *
+   * @param {values} list of objects to generate a color for, usually a
+   * category in a given metadata column.
+   * @param {map} string, name of the discrete color map to use.
+   *
+   * @return {colors} The object containing the hex colors keyed to each sample
```
ElDeveloper: Same comment regarding the `@return` tag.

emperor/support_files/js/color-view-controller.js
```diff
@@ -187,77 +219,162 @@ define([
...
+   *
+   * Retrieve an interpolatd color set.
+   *
+   * @param {values} list of objects to generate a color for, usually a
+   * category in a given metadata column.
+   * @param {map} string, name of the discrete color map to use.
+   *
+   * @return {colors} The object containing the hex colors keyed to each sample
+   *
+   * Defaults to use viridis colormap if there's no map passed in.
```
ElDeveloper: I think this should be part of the `@param {map}`.

emperor/support_files/js/color-view-controller.js
```diff
@@ -187,77 +219,162 @@ define([
...
+   *
+   */
+  ColorViewController.getInterpolatedColors = function(values, map) {
+    map = map || 'viridis';
+    map = chroma.brewer[map];
+
+    var total = values.length;
+    interpolator = chroma.bezier([map[0], map[3], map[4], map[5], map[8]]);
```
ElDeveloper: this is global, I think.

emperor/support_files/js/view-controller.js
```diff
@@ -84,9 +84,9 @@ define([
     this.$header = $('<div name="emperor-view-controller-header"></div>');
     this.$header.css('margin', '0 auto');
     this.$header.css('width', '100%');
-
```
ElDeveloper: Any reason why this line is removed?

emperor/support_files/js/view-controller.js
```diff
@@ -84,9 +84,9 @@ define([
...
     this.$body = $('<div name="emperor-view-controller-body"></div>');
     this.$body.css('margin', '0 auto');
+
```
ElDeveloper: Any reason why this line is added?

emperor/support_files/js/color-view-controller.js
```diff
@@ -174,90 +181,209 @@ define([
...
-   * This function will generate a list of coloring values depending on the
-   * coloring scheme that the system is currently using (discrete or
-   * continuous).
    */
   ColorViewController.getColorList = function(values, map, discrete, scaled) {
-    var colors = {}, numColors = values.length - 1, counter = 0;
-    var interpolator, min, max;
-    var scaled = scaled || false;
+    var colors = {}, gradientSVG;
+    scaled = scaled || false;
```
mortonjt: I'm just curious - do you really need this line?

ElDeveloper:
🚢 :shipit:

emperor/support_files/js/color-view-controller.js
```diff
@@ -174,90 +181,209 @@ define([
...
 
     if (_.findWhere(ColorViewController.Colormaps, {id: map}) === undefined){
       throw new Error("Could not find " + map + " as a colormap.");
     }
 
-    // 1 color and continuous coloring should return the first element of the
-    // map
-    if (numColors === 0 && discrete === false) {
+    // 1 color and continuous coloring should return the first element in map
+    if (values.length == 1 && discrete === false) {
```
mortonjt: `value.length == 1` or `values.length === 1`?

emperor/support_files/js/color-view-controller.js
squirrelo: yes, it makes the coloring default to not showing scaled bar if called, which should happen.

emperor/support_files/js/color-view-controller.js
squirrelo: either works in this case since we are calling length.

emperor/support_files/js/color-view-controller.js
```diff
@@ -174,90 +181,209 @@ define([
...
+   * @param {String} [map] name of the color map to use, see
+   * ColorViewController.Colormaps Defaults to use qiime discrete colors if
+   * there's no map passed in.
+   *
+   * @return {Object} [colors] The object containing the hex colors keyed to
+   * each sample
+   *
+   */
+  ColorViewController.getDiscreteColors = function(values, map) {
+    map = map || 'discrete-coloring-qiime';
```
mortonjt: 👍 pretty slick

emperor/support_files/js/color-view-controller.js
```diff
@@ -174,90 +181,209 @@ define([
...
+
+    if (map == 'discrete-coloring-qiime') {
```
mortonjt: `map == 'discrete-coloring-qiime'` or `map === 'discrete-coloring-qiime'`?

emperor/support_files/js/color-view-controller.js
mortonjt: sounds good.

emperor/support_files/js/color-view-controller.js
mortonjt: gotcha.

PR #469, opened 2016-05-07
ENH: Correctly calculate the size of geometries

ElDeveloper:
We were setting the size of all the geometries to the same hard-coded values,
in the past emperor calculated this in the Python section of the data
processing, the new API allows us to calculate this as part of the scene
creation.


squirrelo:
Just curious, why the change to uppercase for all the shapes? It's a lot of changes for unknown gain.

ElDeveloper:
I'm glad you asked, there are two reasons: (1) the shape names in the dropdown were showing all in lower case (I think we can agree those should be uppercase), and (2) there's no "titlecase" function or method in standard JavaScript, thus making the strings titlecased by default made sense.

emperor/support_files/js/shapes.js
```diff
@@ -3,18 +3,53 @@
...
 
   var $shapesDropdown = $('<select>');
-  for (shape in shapes) {
+  _.each(shapes, function(shape){
     $shapesDropdown.append(new Option(shape, shape));
-  }
+  });
 
-  return {shapes: shapes, $shapesDropdown: $shapesDropdown};
+  return {$shapesDropdown: $shapesDropdown,
```
squirrelo: We should also still return the shapes list so we have a list of what is available.

squirrelo:
One minor comment, 👍 otherwise.

ElDeveloper:
Comment should be addressed!

PR #470, opened 2016-05-12
Render only on changes

squirrelo:
This moves rendering to an event-based action, where new frames are only rendered when the views are changed in some way. previously it was rendering a new frame regardless of it was needed or not.

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -210,8 +210,10 @@ define([
       }
     });
 
-    // Set for update 
-    this.sceneViews[0].needsUpdate = true;
+    //Set all scenes to needing update
+    for (var i = 0; i < this.sceneViews.length, i++) {
+      this.sceneViews[i].needsUpdate = true;
+    }
   };
 
   /**
@@ -404,6 +406,7 @@ define([
         controller.fromJSON(json[index]);
       }
     });
+    sceneview.needsUpdate = true;
    };
 
   /**
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -220,15 +220,16 @@ define([
    *
    **/
   EmperorController.prototype.render = function() {
-    if (!_.some(this.sceneViews, function(sv) {
+    var needsRender = _.some(this.sceneViews, function(sv) {
       return sv.checkUpdate();
-    })) {
-      return;
-    }
-    this.renderer.setViewport(0, 0, this.width, this.height);
-    this.renderer.clear();
-    for (var i = 0; i < this.sceneViews.length; i++) {
-      this.sceneViews[i].render();
+    });
+
+    if (needsRender) {
+      this.renderer.setViewport(0, 0, this.width, this.height);
+      this.renderer.clear();
+      for (var i = 0; i < this.sceneViews.length; i++) {
+        this.sceneViews[i].render();
+      }
     }
   };
 
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -224,7 +224,6 @@ define([
   EmperorController.prototype.render = function() {
     var scope = this;
     $.each(this.sceneViews, function(i, sv) {
-      console.log(i, sv);
       if (sv.checkUpdate()) {
         scope.renderer.setViewport(0, 0, scope.width, scope.height);
         scope.renderer.clear();
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -221,26 +221,14 @@ define([
    **/
   EmperorController.prototype.render = function() {
     if (!_.some(this.sceneViews, function(sv) {
-      // Check if the sceneView or any decView in the sceneView was updated and
-      // needs re-rendering
-      if (_.any(sv.decViews, function(dv) {
-        return dv.needsUpdate;
-      })) {
-        return true;
-      }
-      return sv.needsUpdate;
+      return sv.checkUpdate();
     })) {
-      // Do not render if no changes
       return;
     }
     this.renderer.setViewport(0, 0, this.width, this.height);
     this.renderer.clear();
     for (var i = 0; i < this.sceneViews.length; i++) {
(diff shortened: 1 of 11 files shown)


squirrelo:
No more cooking computer or really loud fan noise[!](https://www.youtube.com/watch?v=4zLfCnGVeL4)

emperor/support_files/js/controller.js
```diff
@@ -209,6 +209,9 @@ define([
         controller.resize(w, h);
       }
     });
+
+    // Set for update 
+    this.sceneViews[0].needsUpdate = true;
```
ElDeveloper: Shouldn't this be a loop over all the sceneViews?

emperor/support_files/js/controller.js
```diff
@@ -217,6 +220,11 @@ define([
    *
    **/
   EmperorController.prototype.render = function() {
+    if (!_.some(this.sceneViews, function(sv) {
```
ElDeveloper: I see what you were talking about before, what about something like this:


```javascript

var shouldRender = _.some(this.sceneViews, function(sv) {
    return sv.checkUpdate();
});

if(shouldRender){
  // rendering shenanigans 
}

```

This way we get rid of the `return` and avoid having the `if` spanning multiple lines.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -92,13 +92,17 @@ define([
     // use get(0) to retrieve the native DOM object
     this.control = new THREE.OrbitControls(this.camera,
                                            $container.get(0));
+    this.control.addEventListener('change', function() {
```
ElDeveloper: 🙌

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -337,10 +341,27 @@ define([
     this.width = width;
     this.height = height;
     this.setCameraAspectRatio(width/height);
+    this.needsUpdate = true;
   };
 
   /**
    *
+   * Convenience method to check if this or any decViews under this need rendering
```
ElDeveloper: if this or any -> if any of the

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -92,13 +92,17 @@ define([
...
+      scope.needsUpdate = true;
+    });
     this.control.rotateSpeed = 1.0;
     this.control.zoomSpeed = 1.2;
     this.control.panSpeed = 0.8;
     this.control.enableZoom = true;
     this.control.enablePan = true;
     this.control.enableDamping = true;
     this.control.dampingFactor = 0.3;
+    this.needsUpdate = true;
```
ElDeveloper: Can you describe this property in the class description docstring?

ElDeveloper:
👍 code looks good! And functionality is great! CPU usage went from 105 % to 3-5% in a 10,000 samples plot.

emperor/support_files/js/controller.js
squirrelo: If any sceneview needs updating all are, so I just used 0 since it will always exist.

emperor/support_files/js/controller.js
squirrelo: Makes sense, I just prefer a hard stop return instead of an if. Done.

emperor/support_files/js/sceneplotview3d.js
squirrelo: 🎉

emperor/support_files/js/sceneplotview3d.js
squirrelo: done

ElDeveloper:
Looks good, all this needs is tests and it should be good to go.

emperor/support_files/js/controller.js
josenavas: Is that always true?

Maybe I missed something but on the design that we did a while ago each scene view was independent so I'm unsure why if 1 needs to change all of them need to.

ElDeveloper: @josenavas, yes you are right, this made me realize that in a situation with more than one `ScenePlotView3D` object, we are updating all the objects if one of them needs update, right?

squirrelo: Yes, the logic on line 223 works exactly like I said above.

squirrelo:
Comments addressed, ready to go.

emperor/support_files/js/controller.js
```diff
@@ -217,11 +222,15 @@ define([
...
    **/
   EmperorController.prototype.render = function() {
-    this.renderer.setViewport(0, 0, this.width, this.height);
-    this.renderer.clear();
-    for (var i = 0; i < this.sceneViews.length; i++) {
-      this.sceneViews[i].render();
-    }
+    var scope = this;
+    $.each(this.sceneViews, function(i, sv) {
+      console.log(i, sv);
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/view-controller.js
```diff
@@ -425,6 +425,8 @@ define([
     var k = this.getActiveDecompViewKey();
     var data = this.decompViewDict[k].setCategory(json.data, this.setPlottableAttributes, json.category);
     this.setSlickGridDataset(data);
+    //set all to needsUpdate
```
ElDeveloper: can you add a leading space in this comment?

emperor/support_files/js/view.js
```diff
@@ -11,8 +11,12 @@ define([
  * @class Contains all the information on how the model is being presented to
  *        the user.
  *
+ * @property {Boolean} [needsUpdate] True when changes have occured that
+ * require re-rendering of the canvas
+ *
  * @param {DecompositionModel} decomp a DecompositionModel object that will be
  * represented on screen.
+
```
ElDeveloper: Can you remove this line?

ElDeveloper:
🎆

colinbrislawn:
I'm so glad this is a thing now. I'm looking forward to the next stable release of Emperor.

PR #472, opened 2016-05-13
API: Remove format.py and make_emperor.py

ElDeveloper:
We no longer need most of format.py, the only function I foresee we will re-use
is the "autograph" creation function, as it has been helpful to debug and
tag created plots. As for make_emperor.py, we will be replacing a lot of the
logic here and use a click command instead.

This marks the first of (hopefully) 3 steps to move the entire project to
Python 3.


ElDeveloper:
This one should be ready to be reviewed, it was a bit more complicated that I initially expected, removes all the code related to make_emperor.py and any trace of qcli. The script to run the tests had to be modified as well and is no longer using qcli, instead it is now using click.

antgonza:
👍

squirrelo:
That is a lot of trimming. Nice.

PR #473, opened 2016-05-16
Info div. Fixes #327

### LINKED ISSUE #327: Click on points to reveal label

squirrelo:
This adds the name of the sample clicked when you click it. 

popup


mortonjt:
Nice!!!

ElDeveloper:
Thanks @squirrelo !

PR #477, opened 2016-05-17
Run linter and doc creation in travis

ElDeveloper:

PR #484, opened 2016-05-18
Render ducks

squirrelo:
This cleans up the documentation pages quite a bit, as well as upgrading the jquery and bootstrap to the newest versions. 

Running `make html` will now render the jsdocs as well as sphinx.

Requires #477


doc/Makefile
```diff
@@ -53,6 +53,7 @@ clean:
 
 html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	jsdoc ../emperor/support_files/js/ -d $(BUILDDIR)/jsdoc -c ../jsdoc-config.json -r
```
ElDeveloper: `jsdoc-config.json` is not committed in this PR. Also if that config file is part of the documentation, I would suggest putting it inside the `doc` directory instead of the base directory.

squirrelo: Correct, the file is added in #477 which this depends on. I can move it over in that PR if you want.

ElDeveloper: Noted, I'm getting confused with all the PRs 😨

squirrelo:
Comments addressed, ready to go.

PR #485, opened 2016-05-21
Scale view controller

squirrelo:
This adds the scale view controller, as well as breaks out the numeric and non-numeric values finding into its own testable util function due to edge cases found.

scaleviewcontroller


jsdoc-config.json
```diff
@@ -0,0 +1,19 @@
+{
```
ElDeveloper: Did the commits get scrambled?

squirrelo: No, file was needed to test compliance and accidentally got committed.

emperor/support_files/js/scale-editor.js
```diff
@@ -0,0 +1,96 @@
...
+      $input = $sliderDiv.slider({
+        range: 'max',
+        min: 0.1,
+        max: 5.0,
+        value: args.item.value,
+        step: 0.1,
+        slide: function(event, ui) {
+          $viewval.val(ui.value);
+          args.item.value = ui.value;
+        }
```
ElDeveloper: Do you think we could subscribe an `onchange`/`changed`/`finshedchanging` callback to this slider so that we don't have to wait to click the grid for the values to change? This can be done similarly to `color-editor`, see this line:
https://github.com/biocore/emperor/blob/new-api/emperor/support_files/js/color-editor.js#L64

The idea behind it, is that once you want to update the value, you just "commit" the changes to the grid, and the visual elements should be updated.

emperor/support_files/js/scale-view-controller.js
```diff
@@ -0,0 +1,278 @@
...
+    /**
+     * jQuery node for checkbox controlling whether to scale by values or not
+     * @type {Node}
+     */
+    this.$scaledValue = $('<input type="checkbox">');
+    /**
+     * jQuery node for label of $scaledValues
+     * @type {Node}
+     */
+    this.$scaledLabel = $('<label for="scaledValue">Scale by values</label>');
```
ElDeveloper: Does label thing need the "for" tag?

emperor/support_files/js/scale-view-controller.js
```diff
@@ -0,0 +1,278 @@
...
+        var data = decompViewDict.setCategory(
+          attributes, scope.setPlottableAttributes, category);
+
+        scope.setSlickGridDataset(data);
+      },
+      'slickGridColumn': {id: 'title', name: '', field: 'value',
+        sortable: false, maxWidth: SLICK_WIDTH,
+        minWidth: SLICK_WIDTH,
+        autoEdit: true,
+        //formatter: Slick.Formatters.Checkmark,
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/scale-view-controller.js
```diff
@@ -0,0 +1,278 @@
...
+  ScaleViewController.prototype.constructor = EmperorAttributeABC;
+
+  /**
+   * Converts the current instance into a JSON string.
+   *
+   * @return {Object} JSON ready representation of self.
+   */
+  ScaleViewController.prototype.toJSON = function() {
+    var json = EmperorAttributeABC.prototype.toJSON.call(this);
+    var sliderVal = this.$globalDiv.children('input').val();
```
ElDeveloper: Minor, but any reason to create this variable to then assign it right below?

emperor/support_files/js/scale-view-controller.js
```diff
@@ -0,0 +1,278 @@
...
+   * Resizes the container and the individual elements.
+   *
+   * Note, the consumer of this class, likely the main controller should call
+   * the resize function any time a resizing event happens.
+   *
+   * @param {float} width the container width.
+   * @param {float} height the container height.
+   */
+  ScaleViewController.prototype.resize = function(width, height) {
+    this.$body.height(this.$canvas.height() - this.$header.height());
```
ElDeveloper: Any reason why the body needs to be resized here? It gets resized again when the `super.resize` is getting called.

emperor/support_files/js/scale-view-controller.js
```diff
@@ -0,0 +1,278 @@
...
+        alert('Non-numeric values detected. These will be hidden!');
+      }
+
+      //scale remaining values between 1 and 5 scale
+      var min = _.min(numeric);
+      var max = _.max(numeric);
+      var range = max - min;
+      _.each(numeric, function(element) {
+          scale[element] = Math.round(
+            (1 + (element - min) * 4 / range) * 10000) / 10000;
```
ElDeveloper: Would you mind adding some comments explaining what's going on here?

emperor/support_files/js/util.js
```diff
@@ -89,6 +89,27 @@ define(['underscore'], function(_) {
 
   /**
    *
+   * Split list of string values into numeric and non-numeric values
+   * @param {String[]} values The values to check
+   * @return {Numeric[]} Array of numeric values, converted to numbers
```
ElDeveloper: Shouldn't this be a list of lists? Or is the convention to use two `@return` tags?

emperor/support_files/js/util.js
```diff
@@ -89,6 +89,27 @@ define(['underscore'], function(_) {
...
+    var nonNumeric = [];
+    _.each(values, function(element) {
+        if (!isNaN(element) && isFinite(+element) && Number(element) !== 0) {
+          numeric.push(+element);
+        }
+        else {
+          nonNumeric.push(element);
+        }
+      });
+    return [numeric, nonNumeric];
```
ElDeveloper: I was thinking about how this function is being used and it may make more sense to return an object instead of the array? Something like:

```javascript
function thyFunction(values){
    // ...
    return {numeric: numeric, nonNumeric: nonNumeric};
}
```

Later the code that uses this can just refer to the appropriate properties in a named manner instead of having to assign these to variables, what do you think?

ElDeveloper:
This works as advertised 👏 !! I made a few comments on the code that should be easy to address. The only thing I noticed: the width of the cell that displays the sliders is taking over the real estate to display the metadata on the right column. It seems like the width could be much smaller, because it mostly displays just a number from 1 to 5. From what I can tell, the reason why it currently is set at that width is because it needs that space for the slider, could we just show a hovering slider instead? Or can you think of an alternative that only ever spans a larger width when needed?

ElDeveloper:
PS - Scale by values is indeed a *strange* feature.

emperor/support_files/js/scale-view-controller.js
squirrelo: Originally it allowed for clicking the label as well to check the box, but since we're not using ids for things I can remove it.

emperor/support_files/js/scale-view-controller.js
squirrelo: Readability and length of line mostly. Squashed to one line now.

emperor/support_files/js/scale-view-controller.js
squirrelo: The issue here is order of ops. the slider needing `this.$globalDiv.width()` means that the body needs to be at the correct size before we do the controller-specific resizing, but the operations in the super need to be done after the controller specific resizing. This allows things to happen in the correct order.

emperor/support_files/js/scale-view-controller.js
squirrelo: yup, done.

emperor/support_files/js/util.js
squirrelo: Makes sense for ease of use. Done.

squirrelo:
I'll try the hover scaling and see how it functions. If it works, I'll commit and add a gif.

squirrelo:
Floating divs added, and here is how they look: 
scaleviewcontroller

emperor/support_files/js/shape-editor.js
```diff
@@ -35,6 +35,11 @@ function($, _, DecompositionView, ViewControllers, shapes) {
     this.init = function() {
       $input = shapes.$shapesDropdown;
       $input.appendTo(args.container);
+      $input.on('change', function() {
+        // commit the changes as soon as a new color is selected
```
ElDeveloper: color -> shape

squirrelo:
Ready to go again

ElDeveloper:
@squirrelo, thanks for making these changes, the code is looking good! So glad we have jshint and everything setup 👍 

I noticed a few glitches though:

1) Slider spans wider than the grid itself (quick guess you may be taking the width of the entire row vs just a column?:

screen shot 2016-06-08 at 10 40 49 pm

2) The global update seems to update size on every change event, I think this needs to be changed so we make sure large datasets don't choke when someone wiggles that slider.

3) When I click to scale numerically, it resets the values of all categories to 1, see the animated GIF below:

rescale

Let me know if you need help, and sorry for the slow response! 🕙

squirrelo:
Good catch on those. I've moved the global update to only change on release of the mouse, so only when a value is set. The scaling by value reset issue is fixed through better handling of the error thrown.

ElDeveloper:
Thanks @squirrelo, everything seems to be working well now, except for the slider width:

screen shot 2016-06-09 at 10 23 16 am

squirrelo:
After speaking with @ElDeveloper looks like this is due to scrollbars being shown when a mouse is plugged in but hidden when no mouse is present, which is something to look out for in future development. Fix is pushed.

PR #497, opened 2016-05-26
DOC: Add docs for color editor & formatter

ElDeveloper:
@squirrelo This is the best I could come up with, let me know what you think.


emperor/support_files/js/color-editor.js
```diff
@@ -9,21 +14,21 @@ function($, _, DecompositionView, ViewControllers, spectrum) {
...
+   * This class represents a color editor defined by the SlickGrid
    * project.
    *
-   * @property {$input} object where the Spectrum color picker lives.
-   * @property {defaultValue} initial value of the cell being edited.
-   *
    * Note, this object is heavily based on classes in slick.editors.js and in
-   * the documentation that can be found here:
-   *    https://github.com/mleibman/SlickGrid/wiki/Writing-custom-cell-editors
+   * the documentation that can be found [here](https://github.com/mleibman/
```
squirrelo: To be more "JSDoc-y", you can change
(https://github.com/mleibman/SlickGrid/wiki/Writing-custom-cell-editors)
 to
{@link https://github.com/mleibman/SlickGrid/wiki/Writing-custom-cell-editors}
and it will still render as a link.

squirrelo:
I like that it properly renders as a class with this setup, but I'm a little hesitant to keep adding more jsdoc exceptions to the linter. We should discuss more tomorrow but overall looks good.

emperor/support_files/js/color-editor.js
ElDeveloper: I tried using the `@link` directive, but since the URL spans multiple
lines, the whitespace messes with the link and leads to a 404. This
problem doesn't seem to happen with the markdown syntax though.

ElDeveloper:
Sounds like a good idea!

emperor/support_files/js/color-editor.js
squirrelo: Makes sense.

PR #498, opened 2016-05-26
STY: Fix tests to fit within 80 columns

ElDeveloper:
Should help with gjslint not passing.


squirrelo:
This is awesome! Travis is being weird about upgrading miniconda, so I pulled down and tested on my machine and it all works. Still want to wait and see if the issue resolves itself, but this is a huge step forward for gjslint compliance!

PR #504, opened 2016-05-26
Duck filled API

squirrelo:
This adds documentation and linting to the javascript, on top of the existing tests.


ElDeveloper:
We are good! We should probably squash all these commits.

squirrelo:
I've squashed some of them. Should be good to go.

ElDeveloper:
Did you forget to push? I still see > 80 commits, would be nice to have 1 or 2 at most, since these are all documentation additions, there's no point in keeping this history, and since you put the majority of work it would make sense for you to author the squashed commit.

squirrelo:
Seems I'm having the sync-ing issue again. I'll try pushing again, but if that doesnt work I'll just close and open another PR.

PR #505, opened 2016-05-26
Last bits

squirrelo:
This fixes the last bits of code needed to get the linter passing and adds a few @see tags to make documentation more helpful.


squirrelo:
And we're green!

PR #508, opened 2016-06-14
Add Python 3 compatibility

### LINKED ISSUE #313: Add support for python 3
- [x] setup travis to test for 2 and 3.
- [x] fix broken tests.
- [x] change make_emperor.py to use click instead of qcli.
- [x] remove unnecessary code in format.py and any dependency from qcli.
- [x] move all_tests.py to click

---

Any takers?

ElDeveloper:
This is still a work in progress. The bulk of the work was done using 2to3 and there were a few hand-made changes, the commit history should describe these in detail.

This also includes a change to .travis.yml to run the PY3 tests.

Should fix #313

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,8 @@ install:
   - npm install -g jsdoc
 script:
   - flake8 emperor/*.py tests/*.py scripts/*.py setup.py
-  - gjslint --custom_jsdoc_tags module,function,constructs,alias,default emperor/support_files/js/*.js tests/javascript_tests/*.js
+  # we can only run gjslint in a python 2.7.x environment
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'; fi
   # execute the full test suite
   - python tests/all_tests.py
   # we just check coverage in the latest version of NumPy
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
 env:
   - NUMPY_VERSION=""  WITH_COVERAGE=1 # environment to test with the latest version of NumPy
   # test only with the latest stable releases of numpy i.e. 1.9.x and 1.10.x
--- a/emperor/qiime_backports/filter.py
+++ b/emperor/qiime_backports/filter.py
@@ -29,8 +29,8 @@ def sample_ids_from_metadata_description(mapping_f,valid_states_str):
     sample_ids = get_sample_ids(map_data, map_header, valid_states)
 
     if len(sample_ids)<1:
-        raise ValueError("All samples have been filtered out for the criteria"+\
-            " described in the valid states")
+        raise ValueError("All samples have been filtered out for the criteria"
+                         " described in the valid states")
 
     return sample_ids
 
@@ -68,7 +68,7 @@ def get_sample_ids(map_data, map_header, states):
     good_ids = []
     for row in map_data:    #remember to exclude header
         include = True
-        for s, vals in list(states.items()):
+        for s, vals in states.items():
             curr_state = row[name_to_col[s]]
             include = include and (curr_state in vals or '*' in vals) \
                 and not '!'+curr_state in vals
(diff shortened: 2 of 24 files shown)


ElDeveloper:
It's alive!

screen shot 2016-06-13 at 6 37 30 pm

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
```
squirrelo: Hooray!

emperor/qiime_backports/filter.py
```diff
@@ -29,8 +29,8 @@ def sample_ids_from_metadata_description(mapping_f,valid_states_str):
     sample_ids = get_sample_ids(map_data, map_header, valid_states)
 
     if len(sample_ids)<1:
-        raise ValueError,"All samples have been filtered out for the criteria"+\
-            " described in the valid states"
+        raise ValueError("All samples have been filtered out for the criteria"+\
```
squirrelo: Don't need the plus here since we have parenthesis now.

emperor/qiime_backports/filter.py
```diff
@@ -68,7 +68,7 @@ def get_sample_ids(map_data, map_header, states):
     good_ids = []
     for row in map_data:    #remember to exclude header
         include = True
-        for s, vals in states.items():
+        for s, vals in list(states.items()):
```
squirrelo: This doesn't need to be a list since it's being iterated anyway.

emperor/qiime_backports/filter.py
```diff
@@ -115,6 +115,6 @@ def filter_mapping_file(map_data, map_header, good_sample_ids,
     headers.append(map_header[-1])
     result.append(to_keep[-1])
     
-    result = map(list,zip(*result))
+    result = list(map(list,list(zip(*result))))
```
squirrelo: actually, if the inner lists don't need to be editable, then you can further simplify to
```python
list(zip(*x))
```

emperor/qiime_backports/parse.py
```diff
@@ -99,11 +96,11 @@ def parse_metadata_state_descriptions(state_string):
     result = {}
     state_string = state_string.strip()
     if state_string:
-        cols = map(strip, state_string.split(';'))
+        cols = list(map(strip, state_string.split(';')))
```
squirrelo: Since this is iterated on in the next line, doesn't need to be changed to a list.

emperor/qiime_backports/parse.py
```diff
@@ -245,8 +240,8 @@ def parse_coords(lines):
         raise QiimeParseError("The line with the vector number was not found"
             ", this information is required in coordinates files")
 
-    lines = map(strip, lines[1:])   #discard first line, which is a label
-    lines = filter(None, lines) #remove any blank lines
+    lines = list(map(strip, lines[1:]))   #discard first line, which is a label
+    lines = [_f for _f in lines if _f] #remove any blank lines
```
squirrelo: Should this line and the next line be combined, or would that reduce readability too much?

emperor/qiime_backports/parse.py
```diff
@@ -263,8 +258,8 @@ def parse_coords(lines):
     #finally, dump the rest of the lines into a table
     header, result = [], []
     for line in lines[:-2]:
-        fields = map(strip, line.split('\t'))
+        fields = list(map(strip, line.split('\t')))
         header.append(fields[0])
-        result.append(map(float, fields[1:]))
+        result.append(list(map(float, fields[1:])))
```
squirrelo: could this block be done as
```python
fields = map(strip, line.split('\t'))
header.append(next(fields))
result.append(lmap(float, fields)))
```
that way the fields list does not have to be created.

emperor/util.py
```diff
@@ -514,18 +513,18 @@ def fill_mapping_field_from_mapping_file(data, headers, values,
...
+        colname, vals = [val for val in v.split(':', 1)]
+        vals = [val.strip() for val in vals.split(',')]
         assert len(vals) == 1, ("You can only pass 1 replacement value:"
                                 " {}".format(vals))
         if colname not in values_dict:
             values_dict[colname] = []
         values_dict[colname].extend(vals)
 
-    for key, v in values_dict.items():
+    for key, v in list(values_dict.items()):
```
squirrelo: Doesn't need the list creation here

tests/_test_core_strings.py
```diff
@@ -12,7 +12,7 @@
 violate PEP-8 rules.
 """
 
-PCOA_STRING = u"""Eigvals	9
+PCOA_STRING = """Eigvals	9
```
squirrelo: I think tests are failing because this u was removed

squirrelo:
Some comments.

squirrelo:
Looks like just travis yml errors

ElDeveloper:
✔️

PR #509, opened 2016-06-27
PY3K: Additional PY3 cleanup

ElDeveloper:
Clean the output for all_tests.py and the descriptors for setup.py.

PR #510, opened 2016-06-28
ENH: Adds changes to support standalone plots

ElDeveloper:
We now have a template that lets you create standalone plots, similar to the
ones that ./build.py would create, instead this time it is part of the main
Emperor API.

I've added the necessary tests, and updated where appropriate.  These changes
are needed to create the Emperor plugin for QIIME 2.


ElDeveloper:
Forgot to mention, this PR depends on #509, so it's better to merge that one first.

emperor/core.py
```diff
@@ -194,15 +200,41 @@ def _repr_html_(self):
...
+        attribute of the object.
+        """
 
-        output.append(style_template.render(base_url=self.base_url))
+        # based on: http://stackoverflow.com/a/6196098
+        loader = FileSystemLoader(join(get_emperor_support_files_dir(),
+                                       'templates'))
+
+        if standalone:
+            main_path = basename(STANDALONE_PATH)
```
antgonza: Should we also copy the resources files? If not, what's the solution for those occasions that we send a plot to a collaborator that has never used Emperor?

tests/_test_core_strings.py
```diff
@@ -67,6 +69,9 @@
     ]);
 }}
 </script>
+
+
```
antgonza: Do we really need these new lines?

tests/_test_core_strings.py
```diff
@@ -188,3 +193,168 @@
 
 }); // END REQUIRE.JS block
 </script>"""
+
+STANDALONE_HTML_STRING = """
```
antgonza: Should these be separate files or are we OK having them here as text? I can see pros/cons for both options ...

tests/_test_core_strings.py
ElDeveloper: No, these are not intentional, those are a consequence of the templates
leaving empty lines after all the `{%....%}` statements.

tests/_test_core_strings.py
ElDeveloper: I agree we can do either or, I decided to do strings because that's what
we already had.

emperor/core.py
ElDeveloper: Great point, I've added this!

tests/_test_core_strings.py
antgonza: K, thanks.

tests/_test_core_strings.py
antgonza: K, was wondering if it will be an issue with pep8 but it seems like we are fine.

ElDeveloper: Got it, yeah, the reason why we are fine is because of this magic line:
https://github.com/biocore/emperor/blob/new-api/tests/_test_core_strings.py#L8

It asks flake8 to ignore all checks inside this file.

emperor/support_files/templates/jupyter-template.html
```diff
@@ -0,0 +1,6 @@
+{# Load templates dynamically, so we can figure out the path on the fly #}
+{% set style_template = style_template_path %}
```
squirrelo: Why set then include?

emperor/support_files/templates/standalone-template.html
```diff
@@ -0,0 +1,30 @@
+{% set style_template = style_template_path %}
```
squirrelo: same question

examples/build.py
```diff
@@ -43,16 +31,26 @@ def listify(a):
...
 coords = np.random.randn(N, 10).tolist()
-pct_var = (1/np.exp(np.arange(10))).tolist()
+pct_var = 1/np.exp(np.arange(10))
+
 
 md_headers = ['SampleID', 'DOB', 'Strings']
 metadata = []
 for _id in coords_ids:
-    metadata.append([_id, ''.join(sample(categories, 1)), ''.join(choice(
+    metadata.append([_id, ''.join(sample(set(categories), 1)), ''.join(choice(
```
squirrelo: out of scope here, but might be good to have a mixed string/character column as well for testing using build.py

squirrelo:
one question, otherwise 👍

ElDeveloper:
I've simplified the templates, thanks!

examples/build.py
antgonza: good point, wanna add an issue?

ElDeveloper:
Thanks both!

PR #512, opened 2016-07-07
Add AxesController [ready for merge]

ElDeveloper:
Most of the functionality for this controller is already present, however
there's still a few details here and there, specifically testing of the
controller is rather shallow.

I'm submitting this to test some things that my computer is flagging as errors
in the JavaScript test suite.


ElDeveloper:
BTW, the interface here is entirely new so feedback is greatly appreciated, everything is centered around a scree plot corresponding to the ordination being displayed, see here:

axes

To test this on your own, I recommend using an IPython notebook or `./examples/build.py`.

mortonjt:
Looking at the error messages, it seems like most of the tests are failing because of Python 2 vs Python 3.

A couple options around this
1. Start using numpy more for testing.  Don't just `assertItemsEquals`, use [`assert_array_equals`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_array_equal.html)
2. Completely ditch Python 2.  I think this will be OK, since all of our tools are moving in that direction anyways.

wasade:
How come the sum of the variation is >100?

ElDeveloper:
@mortonjt Thanks so much for looking at it. Most of the errors were a consequence of #513, all of which should be addressed by #514. As for the suggestion to drop Python 2, this will likely have to be done at some point, but for now the effort to maintain PY2 & PY3 is minimal  `¯\_(ツ)_/¯`

@wasade, I'm using synthetic (read poorly generated) data to do this demo, hence such an abnormality. You can try it in PY2 or PY3 😄

antgonza:
Looks good, is this ready to go?

ElDeveloper:
This is ready assuming #514 is merged, so reviewing/merging that one
first will make the volume of code in this PR much smaller.

ElDeveloper:
Can someone review/merge?

antgonza:
This already has been reviewed but tests are not running ...

ElDeveloper:
Thanks @antgonza, I just pushed a small commit to see if we can get Travis to run, though I think it might just be the case that we have too many tests CI builds running right now ⏰!

Just in case, here's what the final version of this view controller looked like:

axes-3

antgonza:
Looks really cool. Nothing better than fuchsia ...

antgonza:
Actual errors, can you fix?

ElDeveloper:
@antgonza, thanks, just submitted a fix. Also, yes nothing better than "rosa mexicano" 👍

ElDeveloper:
Thanks @antgonza!

Issue #513, opened 2016-07-07
[new-api]: JavaScript tests are not running properly 💀


ElDeveloper:
The JavaScript tests have been executing on Travis incorrectly, however phantomjs doesn't fail and thus doesn't show a failure. I'm looking into solutions right now. 😱

From what I can tell, by running the tests in my browser, most of the tests are passing, but there's a few tests that are broken.

PR #514, opened 2016-07-07
Fix runner.js to execute tests correctly 🎉

ElDeveloper:
The asynchronous nature of require.js lead to a problem where test suites would
not get executed. To fix this, I added all the dependencies to the first
execution of `require`, thus guaranteeing that all resources are available
before the tests begin executing.

As a consequence of all tests not being executed, I found a few failures
(nothing critical); these have now been fixed.

Finally, seems like with recent updates runner.js was producing some error
messages that would make the output of running the tests a bit misleading. I
have also fixed this.

See #513


ElDeveloper:
Can someone have a look at this PR? The changes are rather mechanical but nonetheless important. 🙏

antgonza:
Looks good to me! 👍 Should we merge?

ElDeveloper:
Yes, that would be fantastic!! Thanks for reviewing.

PR #517, opened 2016-07-11
ENH: Add support to plot 🐼s DataFrame objects

ElDeveloper:
This PR adds a helper function, `scatterplot`, that can take as only argument a
Pandas DataFrame and returns an Emperor object. The function itself
extracts the numerical data from the DataFrame and creates a dummy
OrdinationResults object.

Note that this is a rather experimental feature, so testing/suggestions are
very welcome.

This PR depends on #516 so merge that PR first.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -28,7 +28,7 @@ Format Description:
 Citing Emperor
 ==============
 
-Remember to cite Emperor If you use it in any published research, please
+Remember to cite Emperor if you use it in any published research, please
 include the following citation:
 
 .. note::
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -9,26 +9,7 @@ Emperor's Documentation
 Emperor is an interactive next generation tool for the analysis, visualization
 and understanding of high throughput microbial ecology datasets.
 
-This section of the documentation describes the usage of Emperor's main command
-line interface `make_emperor.py`.
-
 The :ref:`genindex` lists the contents of this documentation.
-
-Citing Emperor
-==============
-
-Remember to cite Emperor If you use it in any published research, please
-include the following citation:
-
-.. note::
-	**EMPeror: a tool for visualizing high-throughput microbial community data**. Vazquez-Baeza Y, Pirrung M, Gonzalez A, Knight R. Gigascience. 2013 Nov 26;2(1):16.
-
-
-You can find the `Emperor paper here <http://www.gigasciencejournal.com/content/2/1/16/>`_, and the data presented in this paper can be found `here <http://gigadb.org/dataset/100068>`_.
-
-
-Classes:
-========
 
 .. toctree::
    :maxdepth: 4
@@ -43,3 +24,15 @@ Format Description:
 
    formats
 
+
+Citing Emperor
+==============
+
+Remember to cite Emperor If you use it in any published research, please
+include the following citation:
+
+.. note::
+	**EMPeror: a tool for visualizing high-throughput microbial community data**. Vazquez-Baeza Y, Pirrung M, Gonzalez A, Knight R. Gigascience. 2013 Nov 26;2(1):16.
+
+
+You can find the `Emperor paper here <http://www.gigasciencejournal.com/content/2/1/16/>`_, and the data presented in this paper can be found `here <http://gigadb.org/dataset/100068>`_.
--- a/emperor/__init__.py
+++ b/emperor/__init__.py
@@ -8,5 +8,7 @@
 __version__ = "0.9.51-dev"  # noqa
 
 from emperor.core import Emperor
+from emperor.pandas import scatterplot
 
-__all__ = ['Emperor', 'biplots', 'format', 'filter', 'parse', 'sort', 'util']
+__all__ = ['Emperor', 'scatterplot', 'biplots', 'format', 'filter', 'parse',
+           'sort', 'util']
--- a/emperor/__init__.py
+++ b/emperor/__init__.py
@@ -2,8 +2,8 @@ r"""
 Emperor 3D PCoA viewer (:mod:`emperor`)
 ============================================
 
-This module provides an objects and functions to interact and visualize a set
-of coordinates using the Emperor interface user interface. The `Emperor` class,
+This module provides objects and functions to interact and visualize a set of
+coordinates using the Emperor interface user interface. The `Emperor` class,
 and the functions present here are all intended to be compatible with the
 Jupyter notebook.
 
--- a/emperor/__init__.py
+++ b/emperor/__init__.py
@@ -1,3 +1,29 @@
+r"""
+Emperor 3D PCoA viewer (:mod:`emperor`)
+============================================
+
+This module provides an objects and functions to interact and visualize a set
+of coordinates using the Emperor interface user interface. The `Emperor` class,
+and the functions present here are all intended to be compatible with the
+Jupyter notebook.
+
+.. currentmodule:: emperor
+
+Classes
+-------
+.. autosummary::
+    :toctree: generated/
+
+    Emperor
+
+Functions
+---------
+.. autosummary::
+    :toctree: generated/
+
+    nbinstall
+    scatterplot
+"""
 # ----------------------------------------------------------------------------
 # Copyright (c) 2013--, emperor development team.
 #
@@ -8,7 +34,8 @@
 __version__ = "0.9.51-dev"  # noqa
 
 from emperor.core import Emperor
-from emperor.pandas import scatterplot
+from emperor._pandas import scatterplot
+from emperor.util import nbinstall
 
 __all__ = ['Emperor', 'scatterplot', 'biplots', 'format', 'filter', 'parse',
-           'sort', 'util']
+           'sort', 'util', 'nbinstall']
--- a/emperor/_pandas.py
+++ b/emperor/_pandas.py
@@ -1,19 +1,3 @@
-r"""
-Pandas Interface for Emperor
-============================
-
-This module provides a simple interface to visualize a Pandas DataFrame using
-the Emperor.
-
-.. currentmodule:: emperor.pandas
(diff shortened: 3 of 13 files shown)


emperor/pandas.py
```diff
@@ -0,0 +1,116 @@
...
+def scatterplot(df, x=None, y=None, z=None, remote=True):
+    """Create an Emperor scatter plot from a Pandas DataFrame
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Pandas DataFrame with the data to display
+    x, y, z : str
+        Column names in `df`, these will represent the order of the axes in the
+        final visualization.
```
mortonjt: Can we add a comment here about why these parameters for `x`, `y`, `z` are optional?  I'm assuming that they will take the first 3 columns right?

emperor/pandas.py
```diff
@@ -0,0 +1,116 @@
...
+from skbio import OrdinationResults
+
+
+def scatterplot(df, x=None, y=None, z=None, remote=True):
+    """Create an Emperor scatter plot from a Pandas DataFrame
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Pandas DataFrame with the data to display
```
mortonjt: Alright, I'm a little confused here.

Does this dataframe contain everything?  Metadata and PC coordinates?

I wouldn't think this would be desirable - our OTU tables and metadata are often decoupled, and this would put quite a 
bit of burden on the user to get this sorted out.

In the [canvas biplot code](https://github.com/mortonjt/canvas/blob/master/canvas/biplot.py) we keep all of the metadata and PC coordinates separate.  And I think this greatly enhances useability.

emperor/pandas.py
```diff
@@ -0,0 +1,116 @@
...
+    for col in [z, y, x]:
+        if col is not None:
+            ordered.remove(col)
+            ordered = [col] + ordered
+    samples = samples[ordered].copy()
+
+    # match up the metadata and coordinates
+    df = df.loc[samples.index].copy()
+
+    ores = OrdinationResults(short_method_name='Ax', long_method_name='Axis',
```
mortonjt: What do you think about passing in `short_method_name` and `long_method_name` as optional parameters?

emperor/pandas.py
```diff
@@ -0,0 +1,116 @@
...
+                             eigvals=np.zeros_like(samples.columns),
+                             samples=samples, proportion_explained=variance)
+
+    df.index.name = '#SampleID'
+
+    # HACK: scale the position of the samples to fit better within the screen
+    ores.samples = ores.samples / ores.samples.max(axis=0)
+
+    return Emperor(ores, df, dimensions=8, remote=remote)
```
mortonjt: Any reason why there are `dimensions=8`?

mortonjt:
Thanks for putting this together!  This is a very good start.

emperor/pandas.py
ElDeveloper: @mortonjt, I think I'm guilty for the confusion. Let me try to expand,
and then let's figure out how to make this clearer in the documentation.
What I was trying to do with this function, `scatterplot`, is to provide
a interface for someone to plot environmental data alone i.e. just a
mapping file. Hopefully opening up the user-base to other fields that
don't necessarily work with ordinations. Does this make sense?

emperor/pandas.py
ElDeveloper: Sounds like a good idea, though I'm really not sure if we should even be
exposing/passing these arguments, mainly because this is not intended to
be used as an "ordination" interface ... but now, I don't know, perhaps
the best would be to just not pass anything here(?).

emperor/pandas.py
ElDeveloper: Not a good reason, I chose that arbitrarily, let me fix this.

ElDeveloper:
Thanks so much for reviewing. I've pushed a few changes, let me know
what you think. I'm eager to be easier to read and understand 👓

antgonza:
Can you update and fix error?

emperor/_pandas.py
```diff
@@ -0,0 +1,102 @@
...
+from skbio import OrdinationResults
+
+
+def scatterplot(df, x=None, y=None, z=None, remote=True):
+    """Create an Emperor scatter plot from a Pandas DataFrame
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Pandas DataFrame with the data to display
```
mortonjt: Don't forget to note that this also includes metadata as well as the PC coordinates.

mortonjt:
One comment still needs to be addressed.  But I think this will be ok otherwise.

emperor/_pandas.py
ElDeveloper: Great catch!

ElDeveloper:
I believe I've added all that's needed. I caught a few other kinks when
I built the documentation i.e. `make -C doc html`, but I think I've
addressed them all.

mortonjt:
Cool  👍  on my end.

doc/source/index.rst
```diff
@@ -43,3 +24,15 @@ Format Description:
 
    formats
 
+
+Citing Emperor
+==============
+
+Remember to cite Emperor If you use it in any published research, please
```
antgonza: If -> if

emperor/__init__.py
```diff
@@ -1,3 +1,29 @@
+r"""
+Emperor 3D PCoA viewer (:mod:`emperor`)
+============================================
+
+This module provides an objects and functions to interact and visualize a set
```
antgonza: an objects -> objects?

emperor/_pandas.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def scatterplot(df, x=None, y=None, z=None, remote=True):
+    """Create an Emperor scatter plot from a Pandas DataFrame
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Pandas DataFrame with the data to display, this includes both
+        *metadata* and coordinates to position the samples in a 3D space.
```
antgonza: why *metadata* and not *coordinates*?

emperor/_pandas.py
```diff
@@ -0,0 +1,103 @@
...
+    ValueError
+        If `df` is not a PandasDataFrame
+        If `x`, `y` or `z` are missing from `df` or if they are not numeric
+        columns.
+        If after removing rows with missing data there are fewer than 3
+        samples.
+
+    Notes
+    -----
+    If a row has missing data, that data pont will be removed from the
```
antgonza: pont -> point

emperor/_pandas.py
mortonjt: It contains both metadata and coordinates within the same dataframe.  Right
@ElDeveloper?

On Mon, Jul 11, 2016 at 12:21 PM, Antonio Gonzalez <notifications@github.com

emperor/_pandas.py
```diff
@@ -0,0 +1,103 @@
...
+
+    See Also
+    --------
+    emperor.core.Emperor
+    """
+
+    if not isinstance(df, pd.DataFrame):
+        raise ValueError("The argument is not a Pandas DataFrame")
+
+    for col in [z, y, x]:
```
antgonza: Should you validate that x, y, z actually exist in the df?

antgonza: Just saw the validation below.

emperor/_pandas.py
```diff
@@ -0,0 +1,103 @@
...
+        if col is None:
```
antgonza: How can col be None? Should you be checking for [None, None, None] or any()?

emperor/_pandas.py
```diff
@@ -0,0 +1,103 @@
...
+    # re-order x, y and z
+    ordered = samples.columns.tolist()
+    for col in [z, y, x]:
+        if col is not None:
+            ordered.remove(col)
+            ordered = [col] + ordered
+    samples = samples[ordered].copy()
+
+    # match up the metadata and coordinates
+    df = df.loc[samples.index].copy()
```
antgonza: Out of curiosity, why do you need copy? Basically, what's the difference between `df = df.loc[samples.index].copy()` and `df = df.loc[samples.index]`?

antgonza:
A few comments/questions.

emperor/_pandas.py
ElDeveloper: Fixed @antgonza. @mortonjt, you are correct, though I think @antgonza referred to the formatting of the words, I italicized metadata but I didn't coordinates. 🇮🇹

emperor/_pandas.py
ElDeveloper: These can all be `None`, if they are not explicitly specified, then the columns with the highest variance will take precedence and take the place of `x`, `y` or `z` correspondingly. That's why I chose to make them optional arguments.

antgonza: Got it, thanks. Was reading that line wrongly.

emperor/_pandas.py
ElDeveloper: One represents a "view" of the original data (without using copy) and the other one is a hard copy of the DataFrame (using copy). IIRC, views cannot, or should not, be modified. When you create a copy, you are explicitly creating a new object that's independent of the object from which the frame originated.

ElDeveloper:
Comments should be addressed, let's just wait for the tests to pass.

emperor/_pandas.py
antgonza: That kind of makes sense but when you do: `df = df.loc[samples.index].copy()` it should be the same effect than `df = df.loc[samples.index]`, right? I agree that this will make more sense if it was `df2 = df.loc[samples.index].copy()`.

http://stackoverflow.com/a/27680109

ElDeveloper: You are right! There's no need to make a copy since those two objects
are being overwritten, I've fixed that where needed :+1:.

PR #518, opened 2016-07-12
Add 2 example notebooks 📓 and example data

ElDeveloper:
[The first example](http://nbviewer.jupyter.org/github/ElDeveloper/emperor/blob/200b9f0deae080226777b5038beb2210786b3afd/examples/keyboard.ipynb) notebook, keyboard.ipynb, loads the data from the
keyboard dataset and visualizes the matrix using Emperor.

[The second example](http://nbviewer.jupyter.org/github/ElDeveloper/emperor/blob/200b9f0deae080226777b5038beb2210786b3afd/examples/evident.ipynb) is a bit more complex, as it creates an interactive
Jupyter widget that allows users to select from a variety of different
distances, and rarefaction depths. My inspiration here was to reproduce a
bit of what E-vident was capable of doing, with a few lines of Emperor/skbio/sklearn code.

This animated gif should summarize the second example (it's a bit long, but it's worth it):

evident


antgonza:
Really cool. Next step: add E-vident to Qiita.

- [ ] Anyway, could you add direct links to nbviewer.jupyter.org at the top of each notebook? 
- [ ] Also could you add the study id of the keyboard dataset in Qiita?

ElDeveloper:
Great suggestions, should be there now.

antgonza:
Thanks, could you remove "If you are viewing this notebook, we recommend that you use http://nbviewer.jupyter.org instead of just GitHub, otherwise some functionality will not be available" and put a note beside the Fierer reference that you need to be logged in to Qiita to access it?

ElDeveloper:
Good catch, should be addressed now. 🐍 📓

PR #521, opened 2016-07-14
MAINT: Logistic updates to create Emperor 1.0

ElDeveloper:
These are changes that we will continue to build upon, as we get ready
for 1.0, but in the meantime these are just a start and we can continue
to change these.

Note that these will be used for our pre-release.


ElDeveloper:
# If anyone would like to try this out, you can use `pip install emperor --pre` to download and install the latest "beta" release of emperor.

PR #523, opened 2016-07-17
ENH: Adds the ability to use autorotation

ElDeveloper:
Users can now click on the context menu to toggle autorotation. In doing this,
I also improved the icons that we were using in the context menu i.e. save and
open now have meaningful icons.

cc @wdwvt1 (a long time ago you mentioned this would be useful :tada:)


ElDeveloper:
Small demo:

rotate

antgonza:
Looks great, thanks! I'm not sure but thought on raising the question, should we have a left/right rotation?

ElDeveloper:
Maybe(?), seems like in `ili there's only an option to have right
rotation. Which reminds me, a few members of the Dorrenstein group
requested this feature, so I guess only right may be OK?

antgonza:
K, if we can always add on a later date after users request it. Thanks!

ElDeveloper:
Yes, that's what I was thinking, thanks!!

PR #525, opened 2016-07-27
Flipping bug 🐞

ElDeveloper:
In the current version of emperor, if you flip the orientation of one of the axis, you may end up in a situation where the range data lies outside of the current axes. To remediate this, the `dimensionRanges` properties are updated to reflect the new range that the data spans. Additionally, we need to keep track of when an axis is reoriented, so we know when change the ranges back in a situation when we are swapping out a *flipped* dimension.

I've also improved the test cases to reproduce the problem.

-------

This one took longer than I was hoping. 😫

### DIFF
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -266,7 +266,10 @@ define([
       // update the dimensions that changed.
 
       // if we have to compute the data, clean up the previously known ranges
-      this.dimensionRanges = {'min': [], 'max': []};
+      this.dimensionRanges.max = [];
+      this.dimensionRanges.max.length = 0;
+      this.dimensionRanges.min = [];
+      this.dimensionRanges.min.length = 0;
     }
 
     _.each(this.decViews, function(decView, name) {
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -264,6 +264,7 @@ define([
     else {
       // TODO: If this entire function ever becomes a bottleneck we should only
       // update the dimensions that changed.
+      // See: https://github.com/biocore/emperor/issues/526
 
       // if we have to compute the data, clean up the previously known ranges
       this.dimensionRanges.max = [];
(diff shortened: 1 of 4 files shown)


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -225,12 +225,52 @@ define([
...
 
-    // means we already have the data, so let's say goodbye
-    if (this.dimensionRanges.max.length !== 0) {
+    if (computeRanges === false) {
+      // If at this point we still don't need to compute the data, it is safe
+      // to exit because all data still exists within the expected ranges
       return;
     }
+    else {
+      // TODO: If this entire function ever becomes a bottleneck we should only
```
antgonza: How will we know that it became a bottleneck? If a TODO, should we create an issue?

antgonza:
One question, then 👍

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: Great question! My answer is I don't have a good feeling for it, but I think it's not a big deal right now. For the time being I've opened https://github.com/biocore/emperor/issues/526 and made a note in the source.

emperor/support_files/js/view.js
```diff
@@ -41,6 +41,13 @@ function DecompositionView(decomp) {
    */
   this.visibleDimensions = [0, 1, 2]; // We default to the first three PCs
   /**
+   * Orientation of the axes, `-1` means the axis is flipped, `1` means the
+   * axis is not flipped.
+   * @type {integer[]}
+   * @default [1, 1, 1]
+   */
+  this.axesOrientation = [1, 1, 1];
```
josenavas: Probably I'm missing something, but can this be a boolean? looking at the code in here it doesn't look like you're doing any computation with them.

josenavas: Ooohh nevermind - I just saw when they're used!

ElDeveloper: Thanks. Just for completeness, I also thought about that, but it was
somewhat more innatural.

ElDeveloper:
Thanks @josenavas!!

PR #528, opened 2016-07-27
Surgically add the EmperorViewController class

ElDeveloper:
After starting the implementation of `AnimationsViewController`, I noticed that there would be a fair amount of repeated code between this new class, `AxesController` and `EmperorAttributeABC`, therefore I decided to move that redundant code to a new base class `EmperorViewController`.

This new class understands what a decomposition view dictionary is, but doesn't really alter the functionality that EmperorViewControllerABC provides. The tests and most of the code were extracted from `EmperorAttributeABC`.

### DIFF
--- a/emperor/support_files/js/axes-controller.js
+++ b/emperor/support_files/js/axes-controller.js
@@ -8,7 +8,7 @@ define([
 ], function($, _, DecompositionView, ViewControllers, d3, contextmenu) {
 
   // we only use the base attribute class, no need to get the base class
-  var EmperorViewControllerABC = ViewControllers.EmperorViewControllerABC;
+  var EmperorViewController = ViewControllers.EmperorViewController;
 
   /**
    * @class AxesController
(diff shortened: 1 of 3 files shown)


antgonza:
I think it looks fine 👍 . @ElDeveloper do we need another reviewer?

ElDeveloper:
Another reviewer would be OK with me, but perhaps not necessary.

emperor/support_files/js/view-controller.js
```diff
@@ -253,6 +239,75 @@ define([
...
+    return this.activeViewKey;
+  };
+
+  /**
+   * Changes the metadata column name to control.
+   *
+   * @param {String} k Key corresponding to active decomposition view.
+   */
+  EmperorViewController.prototype.setActiveDecompViewKey = function(k) {
+    // FIXME: this should be validated against decompViewDict i.e. we should be
```
antgonza: Just noticed this, should this be an issue?

wasade: I can look through in a moment

emperor/support_files/js/axes-controller.js
```diff
@@ -8,7 +8,7 @@ define([
 ], function($, _, DecompositionView, ViewControllers, d3, contextmenu) {
 
   // we only use the base attribute class, no need to get the base class
```
wasade: is this comment still accurate?

emperor/support_files/js/view-controller.js
wasade: sorry, this comment got placed in the wrong place, meant i can look through as a second reviewer

tests/javascript_tests/test_view_controller.js
```diff
@@ -133,6 +134,117 @@ requirejs([
...
+          this.sharedDecompositionViewDict);
+      equal(attr.activeViewKey, 'pcoa');
+    });
+
+    /**
+     *
+     * Tests to make sure the exceptions are being raised as expected
+     *
+     */
+    test('Cosnstructor test exceptions', function(assert) {
```
wasade: Constructor

wasade:
Minor comments, :+1: otherwise

emperor/support_files/js/view-controller.js
```diff
@@ -253,6 +239,75 @@ define([
...
+    // verifying that the key indeed exists
+    this.activeViewKey = k;
+  };
+
+  /**
+   *
+   * @class EmperorViewControllerABC
```
mortonjt: Update?

mortonjt:
Solid.  👍

emperor/support_files/js/axes-controller.js
ElDeveloper: You are right, this is out of date.

ElDeveloper:
Thanks all for the review. All comments should be addressed. @antgonza,
I didn't open a new issue, I just fixed the problem :) and added a test
to check the exception is raised. Aside from that all the other comments
were also addressed.

ElDeveloper:
Thanks everyone!

PR #532, opened 2016-09-06
Transformer

wasade:
@mortonjt @ElDeveloper


ElDeveloper:
@wasade, when you showed us the cube-looking cloud of points, what transform were you using? The zscore or the min_max? I'm trying, if you can plugging this in with a "real" dataset, might be easier to understand than with the random data generated by build.py. An easy way would be to use the Jupyter notebook in the examples folder.

wasade:
flake8 :( will fix

was using min/max and directly on the `randn` data in `build.py`, I just pushed in a change which recreates it

ElDeveloper:
Thanks @wasade, after thinking more about this problem, I noticed that we want something like scikit-learn's [`scale` function](https://github.com/scikit-learn/scikit-learn/blob/51a765a/sklearn/preprocessing/data.py#L79), except we would want to preserve the relative length of the axes. What I mean by this is that if PC1 spans twice as many units as PC2, then we would want to preserve this ratio, so I think the scaling has to be done iteratively on an axis by axis basis (however, I don't have a proof to support this claim).

When I just use the scale function, the ratios I mentioned before are lost, for example see the comparison in this notebook:

http://nbviewer.jupyter.org/gist/ElDeveloper/13289d380c36c6d7efb6e43bfba263fb

We are using the keyboard dataset, in the first plot we show the original, while in the second we show the scaled, note how the relative position of the points is distorted in the second figure.

wasade:
Oh, nice, yes, lets definitely leverage an existing method if possible. The
intent of the current PR was to operate on each axis individually but it's
possible that isn't happening. Wouldn't the ratio of the differences still
be exposed via % explained?

ElDeveloper:
@wasade, it would be exposed numerically but not visually, which is my biggest concern here. @antgonza any thoughts?

wasade:
2,3, or 4 work for you today?

ElDeveloper:
Yes, why don't we do 2 today?

Thanks!

Yoshiki.

wasade:
Confirmated

ElDeveloper:
OK, today @wasade and I met to discuss, and we found that scikit-learn's `scale` works well enough. But that it also distorts the spacing of the data a little bit.

This notebook should show the comparison:

http://nbviewer.jupyter.org/gist/ElDeveloper/d4b64188ee27dadc80f96e9d61a991c2

-------------

And in case something happens to the notebook, this is the other way to look at it:

## Original

screen shot 2016-10-10 at 5 10 30 pm


## Scaled

screen shot 2016-10-10 at 5 12 18 pm

ElDeveloper:
For completeness, this is an example of a plot where you can see some artifacts show up due to the "numerical instability":

screen shot 2016-10-11 at 9 34 37 am

rob-knight:
do we really have to use custom polyhedra rather than spheres as primitives? seems like the polyhedra approach causes us a lot of trouble?

On Oct 11, 2016, at 9:35 AM, Yoshiki Vázquez Baeza <notifications@github.com<mailto:notifications@github.com>> wrote:


For completeness, this is an example of a plot where you can see some artifacts show up due to the "numerical instability":

[screen shot 2016-10-11 at 9 34 37 am]<https://cloud.githubusercontent.com/assets/375307/19279353/f3531334-8f95-11e6-80bb-23a54e2498dd.png>

wasade:
@rob-knight we believe the range of each axis is causing clipping issues with OpenGL stemming from inherently single precision floats, the matrix multiplications performed by the rendering engine and values which are extremely large (i.e., in the millions) and some are extremely small (i.e., < 0.001). Once we scale the extremes, the clipping goes away. As per the geometry, it's using the [sphere primitives](https://github.com/biocore/emperor/blob/46625a6ba1d37c665198c46adc93ec8d4a7aa734/emperor/support_files/js/shapes.js#L33) from threejs

rob-knight:
ok thanks for clarification

On Oct 11, 2016, at 1:47 PM, Daniel McDonald <notifications@github.com<mailto:notifications@github.com>> wrote:


@rob-knight<https://github.com/rob-knight> we believe the range of each axis is causing clipping issues with OpenGL stemming from inherently single precision floats, the matrix multiplications performed by the rendering engine and values which are extremely large (i.e., in the millions) and some are extremely small (i.e., < 0.001). Once we scale the extremes, the clipping goes away. As per the geometry, it's using the sphere primitives<https://github.com/biocore/emperor/blob/46625a6ba1d37c665198c46adc93ec8d4a7aa734/emperor/support_files/js/shapes.js#L33> from threejs

ElDeveloper:
@wasade, after some light experimentation, I think the scaling can be further simplified if we just divide the entire matrix by the maximum element of the absolute value of the original matrix i.e. something like:

```python
coords \= np.max(np.abs(coords))
```

wasade:
:shipit:

ElDeveloper:
@wasade would you be interested in putting that change in place? 🙏

wasade:
When are you hoping for it?

ElDeveloper:
Any time within the next two weeks?

ElDeveloper:
This PR has been superseded by #570

PR #533, opened 2016-09-06
generating SVG

antgonza:
Missing:
- [x] Fix labels issues
- [x] Do tests pass?

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -460,7 +460,7 @@ define([
       if (this.dm.length >= 9000) {
         if (confirm('This number of samples could take a long time and in ' +
            'some computers the browser will crash. If this happens we ' +
-           'suggest to use the png implementation. Do you want to '
+           'suggest to use the png implementation. Do you want to ' +
            'continue?') == false) {
           return;
         }
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -10,10 +10,13 @@ define([
     'shapecontroller',
     'axescontroller',
     'scaleviewcontroller',
-    'filesaver'
+    'filesaver',
+    'svgrenderer',
+    'draw'
 ], function($, _, contextMenu, THREE, DecompositionView, ScenePlotView3D,
              ColorViewController, VisibilityController, ShapeController,
-             AxesController, ScaleViewController, FileSaver) {
+             AxesController, ScaleViewController, FileSaver, SVGRenderer,
+             Draw) {
 
   /**
    *
@@ -402,13 +405,24 @@ define([
           }
         },
         'sep1': '---------',
-        'saveImage': {
-          name: 'Save Image (PNG)',
-          icon: 'file-picture-o',
-          callback: function(key, opts) {
-            scope.screenshot();
-          }
-        }
+        "fold1": {
+            "name": "Save Image",
+            icon: 'file-picture-o',
+            "items": {
+              'saveImagePNG': {
+                name: 'PNG',
+                callback: function(key, opts) {
+                  scope.screenshot('png');
+                }
+              },
+              'saveImageSVG': {
+                name: 'SVG + labels',
+                callback: function(key, opts) {
+                  scope.screenshot('svg');
+                }
+              }
+            }
+        },
       }
     });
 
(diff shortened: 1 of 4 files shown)


ElDeveloper:
This works as advertised, but the problem with the labels doesn't seem to have an easy solution. Just in case this helps, we are rendering the text as sprites, because this is the easiest way to render text. As such when [we create the sprite](https://github.com/biocore/emperor/blob/new-api/emperor/support_files/js/draw.js#L96), you also need to create a "canvas" where the texture is drawn, the texture in our case is rasterized text. When SVGRender is rendering, the object goes over all the rendered elements in the scene and does something appropriate for each of them. In [this patch](https://github.com/gwindes/three.js/commit/df45c5ed071029b7484ced5f99647727b30ccd44) (courtesy of @antgonza) you can see how textures get rendered simply as rects, regardless of their textures, hence we see the squares overlaid on the plot.

I can think of a couple solutions:

- Fix SVGRenderer to properly retrieve the texture and render it.
- Hack SVGRenderer/Emperor to search for those squares and replace them for labels.
- Render the canvas as a high-resolution PNG, then use JSPDF to shove the data into a PDF, the problem here is that we wouldn't have a vector image, but then again it would be more manipulatable than the current SVGs (see this [example](http://jsfiddle.net/p4s5k59s/1222/), courtesy of this [SO Answer](http://stackoverflow.com/questions/19699366/html5-canvas-to-pdf)).

Any ideas?

antgonza:
I don't like the png->pdf solution as it's not editable. I think SVGRender fix/hack is the way to go. What about doing a fix and if it's not merged quickly we do the hack? -- basically same code.

ElDeveloper:
@antgonza Sounds good, what about staging the changes in Emperor i.e fixing our copy of SVGRenderer, and if we see it is working well then we contribute back to three.js? At least that way we would not wait on their reviews, etc (in case that's a concern).

antgonza:
This is ready for one more pass and should be working fine.

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+      var names = [], colors = [];
+      _.each(this.controllers.color.bodyGrid.getData(), function(element, idx) {
+        names.push(element.category);
+        colors.push(element.value);
+      });
+      var blob = new Blob([Draw.formatSVGLegend(names, colors)],
+                          {type: 'image/svg+xml'});
+      saveAs(blob, 'emperor-image-labels.svg');
+    } else {
+      console.log('Screenshot type not implemented');
```
ElDeveloper: Can you make this a console.error?

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+      var download = $('<a href="' + c + '" download="emperor.' + type + '">');
+      download.get(0).click();
+    } else if (type === 'svg') {
+      // confirm box based on number of samples: better safe than sorry
+      if (this.dm.length >= 9000) {
+        if (confirm('This number of samples could take a long time and in ' +
+           'some computers the browser will crash. If this happens we ' +
+           'suggest to lower the number of segments or use the png ' +
+           'implementation. Do you want to continue?') == false) {
+              return;
```
ElDeveloper: I think this line is over-indented, can you bring it back a few spaces?

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+                       this.renderer.getClearColor().getHexString() + '"/>';
+      index = svgfile.indexOf('>', index) + 1;
+      svgfile = svgfile.substr(0, index) + background + svgfile.substr(index);
+
+      var blob = new Blob([svgfile], {type: 'image/svg+xml'});
+      saveAs(blob, 'emperor-image.svg');
+
+      // generating legend
+      var names = [], colors = [];
+      _.each(this.controllers.color.bodyGrid.getData(), function(element, idx) {
```
ElDeveloper: We don't need the idx parameter in this function.

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+        if (confirm('This number of samples could take a long time and in ' +
+           'some computers the browser will crash. If this happens we ' +
+           'suggest to lower the number of segments or use the png ' +
+           'implementation. Do you want to continue?') == false) {
+              return;
+        }
+      }
+
+      // generating SVG image
+      var svgRenderer = new THREE.SVGRenderer({ antialias: true,
```
ElDeveloper: can you remove the leading space in antialias?

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+    if (type === 'png') {
+      // Render all scenes so it's rendered in same context as save
+      for (var i = 0; i < this.sceneViews.length; i++) {
+        this.sceneViews[i].render();
+      }
+      var c = this.renderer.domElement.toDataURL('image/' + type);
+      // Create DOM-less download link and click it to start download
+      var download = $('<a href="' + c + '" download="emperor.' + type + '">');
+      download.get(0).click();
+    } else if (type === 'svg') {
```
ElDeveloper: Now that we have the option to save as SVG, can you document this in the docstring for this method?

emperor/support_files/js/controller.js
```diff
@@ -431,14 +445,76 @@ define([
...
+      // confirm box based on number of samples: better safe than sorry
+      if (this.dm.length >= 9000) {
+        if (confirm('This number of samples could take a long time and in ' +
+           'some computers the browser will crash. If this happens we ' +
+           'suggest to lower the number of segments or use the png ' +
```
ElDeveloper: We no longer have the ability to decrease the number of segments, can you remove that from this message?

emperor/support_files/vendor/js/three.js-plugins/SVGRenderer.js
```diff
@@ -159,7 +159,15 @@ THREE.SVGRenderer = function () {
 				_v1 = element;
 				_v1.x *= _svgWidthHalf; _v1.y *= - _svgHeightHalf;
 
-				renderSprite( _v1, element, material );
+				if ('text' in element.object) {
```
ElDeveloper: Would you mind making a note of these changes at the top of the file? Just to keep track of this.

ElDeveloper:
Fantastic, tested this and it works as expected!!! ✨ I added a few minor comments, it should be good to go after those are addressed.

ElDeveloper:
Fantastic, thanks @antgonza!

PR #536, opened 2016-09-16
Add the ability to control multiple decomposition views

ElDeveloper:
The main class, `EmperorController`, now has the ability to add additional `DecompositionView` objects. When a new object is added, a message is propagated to the controllers to refresh their metadata categories to include any new information.

Of note is that the metadata menu for controllers now include groups with the name of the decomposition view, and under that group, the metadata categories for that object.

These changes are a bit abstract, but I've commented (hopefully to make as readable as possible) the most esoteric sections.


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -219,6 +214,28 @@ define([
...
+
+    // Add all the meshes to the scene, iterate through all keys in
+    // decomposition view dictionary
+    for (var decViewName in this.decViews) {
+      for (var j = 0; j < this.decViews[decViewName].markers.length; j++) {
+        this.scene.add(this.decViews[decViewName].markers[j]);
+      }
+    }
+
+    this.needsUpdate = true;
```
wasade: doesn't a `refresh` imply any update necessary has already happened?

wasade:
weird... the new gh review system. that'll take a second to get used too

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: Good point, I've renamed the method to something clearer.  Note that we use the `needsUpdate` property to know when we have to render a new frame. Hence in this case when we add the objects to the scene, we notify the renderer that a new frame should be drawn on screen.

ElDeveloper:
This reviews are certainly strange :L

emperor/support_files/js/axes-controller.js
```diff
@@ -128,7 +128,7 @@ define([
       this.$table.remove();
     }
 
-    var view = this.decompViewDict[this.activeViewKey], scope = this;
+    var view = this.getView(), scope = this;
```
mortonjt: :+1: this is already looking so much cleaner!

emperor/support_files/js/view-controller.js
```diff
@@ -247,35 +241,15 @@ define([
...
-   * Retrieves the currently active decomposition view.
+   * This class does not operate on single decomposition views, hence this
+   * method retrieves the first available view.
    *
-   * @return {DecompositionView} The currently active decomposition view.
    */
-  EmperorViewController.prototype.getActiveView = function() {
-    return this.decompViewDict[this.getActiveDecompViewKey()];
+  EmperorViewController.prototype.getView = function() {
+    return this.decompViewDict[Object.keys(this.decompViewDict)[0]];
```
mortonjt: Remember, this was temporary, since this only grabs the first key.  This will need to be changed for biplots. This doesn't need to be fixed here,  but we should add a todo, or an issue so we don't forget that [0] needs to be replaced with a var

emperor/support_files/js/view-controller.js
```diff
@@ -488,16 +516,59 @@ define([
...
+   *
+   * Performs some additional logic to avoid duplicating decomposition names.
+   *
+   * Note that decompositions won't be updated if they have the same name and
+   * same metadata headers, if the only things changing are coordinates, or
+   * metadata values, the changes should be performed directly on the objects
+   * themselves.
+   *
+   */
+  EmperorAttributeABC.prototype.refreshMetadata = function() {
```
mortonjt: Just to clarify, this refreshing all of the metadata associated with the current view? 

Remember, every decomposition view will have their own set of metadata.  

Samples -> sample metadata (ex ph, location)
Features -> feature metadata (ex taxonomy )

If not,  then this will need to be clarified in the docs, along with a todo/issue.

tests/javascript_tests/test_view_controller.js
```diff
@@ -438,5 +448,38 @@ requirejs([
...
       });
     });
+
+
+    /**
+     *
+     * Test refresh metadata.
+     *
+     */
+    asyncTest('Test refreshMetadata', function() {
```
mortonjt: I'm just curious, when do you anticipate using refreshMetadata?  Do you expect to pass in additional metadata during runtime?  If so,  some more detail on this would be nice to include in the docs.

emperor/support_files/js/view-controller.js
ElDeveloper: You are right, this was a temporary fix, but in this case, this method is for the intermediate EmperorViewController class, this class does not operate on a single view at any time, it does operations in bulk, however it is useful to get the zeroth view (as a reference). The controller that will be used for biplots for example, will be EmperorAttributeViewController, and in that one we can select which view we want.

emperor/support_files/js/view-controller.js
ElDeveloper: Samples and features both are views, so you are right, this updates them both.

tests/javascript_tests/test_view_controller.js
ElDeveloper: Yes, during run time, for example with the animations.

mortonjt:
I think this is looking ok.  @josenavas @antgonza @wasade want to merge this?

emperor/support_files/js/view-controller.js
antgonza: Can you add a note with that explanation in the code?

mortonjt:
@ElDeveloper is this ready to merge?

ElDeveloper:
Yes, all comments have been addressed.

mortonjt:
Yay!  Thanks @ElDeveloper !

ElDeveloper:
Great, thanks all for the reviews and merging!

PR #537, opened 2016-09-30
ENH: Add support for custom axes

ElDeveloper:
This addition resuscitates the old implementation from
make_3d_plots.py/make_emperor.py to deal with 'custom axes', however moving
forward we hope that this feature will be available from the user interface.

cc @gregcaporaso


ElDeveloper:
Related to https://github.com/qiime2/q2-emperor/pull/12

gregcaporaso:
Tested locally and it seems to be working. I haven't been able to look at the code yet.

ElDeveloper:
@antgonza, @josenavas, @mortonjt, @wasade, anyone has time to review this PR?

emperor/core.py
```diff
@@ -278,6 +291,31 @@ def make_emperor(self, standalone=False):
         # avoid unicode strings
         coord_ids = list(map(str, self.mf.index.tolist()))
 
+        # TODO: This will be removed once the custom axes creation is moved to
+        # the graphical user interface i.e. to the Axes tab.
```
mortonjt: We really need to create issues for these.  These easter eggs will be forgotten in the future.

emperor/core.py
```diff
@@ -278,6 +291,31 @@ def make_emperor(self, standalone=False):
         # avoid unicode strings
         coord_ids = list(map(str, self.mf.index.tolist()))
 
+        # TODO: This will be removed once the custom axes creation is moved to
```
wasade: Should this be an issue? or prefer to leave as TODO?

tests/test_core.py
```diff
@@ -126,5 +126,27 @@ def test_copy_support_files_use_target(self):
...
+        with open('test.txt', 'w') as f:
+            f.write(obs)
+
+        try:
+            self.assertItemsEqual(tcs.HTML_STRING_CUSTOM_AXES.split('\n'),
+                                  obs.split('\n'))
+        except AttributeError:
+            self.assertCountEqual(tcs.HTML_STRING_CUSTOM_AXES.split('\n'),
+                                  obs.split('\n'))
+        self.assertEqual(tcs.HTML_STRING_CUSTOM_AXES, obs)
```
mortonjt: Why the try/except inside this test?  Can we create 2 tests instead?

emperor/core.py
```diff
@@ -278,6 +291,31 @@ def make_emperor(self, standalone=False):
...
+        # the graphical user interface i.e. to the Axes tab.
+        if custom_axes:
```
antgonza: Should we add a validation that no more than 2 axes are passed?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -397,7 +397,12 @@ define([
           text = decomp.axesNames[index];
         }
       }
-      text += ' (' + decomp.percExpl[index].toPrecision(4) + ' %)';
+
+      // account for custom axes (their percentage explained will be -1 to
+      // indicate that this attribute is not meaningful).
+      if (decomp.percExpl[index] >= 0) {
```
antgonza: What about N/A or even better the name of the column used? -1 is an value and perhaps will confuse users. Same when using more than 1 axis.

emperor/util.py
```diff
@@ -397,8 +398,9 @@ def preprocess_coords_file(coords_header, coords_data, coords_eigenvals,
 
             # this opens support for as many custom axes as needed
             axes = len(custom_axes)
-            coords_low[:, 0:axes] = zeros([coords_low.shape[0], axes])
-            coords_high[:, 0:axes] = ones([coords_high.shape[0], axes])*0.00001
+            coords_low[:, 0:axes] = np.zeros([coords_low.shape[0], axes])
+            coords_high[:, 0:axes] = (np.ones([coords_high.shape[0], axes])
```
wasade: What about `np.full`?

antgonza:
Tests also failed 😏

emperor/util.py
```diff
@@ -652,3 +654,46 @@ def guess_coordinates_files(dir_path):
...
+        `custom_axes`.
+    """
+    # avoid side-effects
+    mf = mf.copy()
+
+    headers = [mf.index.name] + mf.columns.tolist()
+    missing_headers = set(custom_axes).difference(set(headers))
+
+    if missing_headers:
+        raise KeyError("One or more headers are not present in the "
```
wasade: Isn't this situation, one or more requested custom axes is not present in the sample metadata?

emperor/util.py
```diff
@@ -652,3 +654,46 @@ def guess_coordinates_files(dir_path):
...
+                       "sample information: %s" % ', '.join(missing_headers))
+
+    for axis in custom_axes:
+        mf[axis] = pd.to_numeric(mf[axis], errors='coerce')
+
+        if np.any(mf[axis].apply(np.isnan)):
+            raise ValueError("The column '%s', has non numeric values. All "
```
wasade: Do you think it would be useful to provide an example sample and value?

tests/_test_core_strings.py
```diff
@@ -376,6 +376,155 @@
...
+              'slickeditors']
+}
+});
+
+requirejs(
+["jquery", "model", "controller"],
+function($, model, EmperorController) {
+  var DecompositionModel = model.DecompositionModel;
+
+  var div = $('#emperor-notebook-0x9cb72f54');
```
wasade: is this magical hex stable?

wasade: ...or does it need to be? assuming not but thought i'd ask

emperor/util.py
```diff
@@ -652,3 +654,46 @@ def guess_coordinates_files(dir_path):
...
+
+    headers = [mf.index.name] + mf.columns.tolist()
+    missing_headers = set(custom_axes).difference(set(headers))
+
+    if missing_headers:
+        raise KeyError("One or more headers are not present in the "
+                       "sample information: %s" % ', '.join(missing_headers))
+
+    for axis in custom_axes:
+        mf[axis] = pd.to_numeric(mf[axis], errors='coerce')
```
wasade: This might be a pain to support and may be considered out of scope, but it might be cool to allow lexicographic sort. It would also allow for the case where the sample values are like `{"01.daily", "02.weekly", "03.monthly", "04.biannually", "05.annually"}` where the advantage is that the values associated with the metadata have meaning while using `{1..5}` in this context would potentially be confusing

emperor/core.py
ElDeveloper: Opened an issue: https://github.com/biocore/emperor/issues/538

tests/test_core.py
ElDeveloper: The try/except is to deal with Python2 vs Python3. If you guys think it makes more sense to separate the tests one for py2 and one for py3, then I can do that.

emperor/core.py
ElDeveloper: I've opened a new issue: https://github.com/biocore/emperor/issues/538

emperor/util.py
ElDeveloper: Good idea, I will use that instead.

emperor/util.py
ElDeveloper: Yes, you are correct, will fix this.

emperor/util.py
ElDeveloper: Yes, we've talked about this a fair bit, what would really be awesome is if the custom axes could infer other data types like timestamps, but for the time being we'll leave it as numbers only. Once we have something like #538 we should build the different sorting on top of that.

emperor/util.py
ElDeveloper: Great idea, this has now been added.

tests/_test_core_strings.py
ElDeveloper: Yes, this is a magical and stable hexadecimal number. Some background, this is the way that we "guarantee" all div identifiers in a DOM will be unique we randomly generate this string and set it as the ID for the div. If you look at the setUp method, you will see np.seed 🌱

emperor/core.py
ElDeveloper: I don't think we ever limited the number of custom axes that we have. Is there a reason why we should do that? I may be forgetting a technical limitation.

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: Ah, I wanted to use N/As but then the formatting code in the jinja template would coerce the values to zeros, that's why I ended up using a negative value. There should never be negative percentage of explained variances, that's why I went with this.

emperor/core.py
mortonjt: Awesome!

tests/test_core.py
mortonjt: The other option is to completely drop PY2 support.  There is no reason to keep that anymore.

emperor/core.py
antgonza: 1 => 1 vs. PC1 vs. PC2, 
2 => 1 vs. 2 vs. PC1,
3 => 1 vs. 2 vs. 3 -- we don't need PCs 
4 => 1 vs. 2 vs. 3 -- we don't need PCs and 4 ...

ElDeveloper:
> Also, keep in mind that there is no law that says that we need to support backwards compatible changes. For instance, I have not seen any usage of vectors over the last several months. This is an opportunity to boot out some of the less commonly used features.

I am down to remove features, so as long as we discuss this as a group 👍  FWIW, I used the vectors in a figure for a recent manuscript.

mortonjt:
👍 sounds good.  And yeah, I agree this is out of the scope of this PR.

This PR I think is good to go.  But I see that there are still some comments from @antgonza ...

emperor/core.py
ElDeveloper: You are right, but this should be fine, since we can change the axes with the Axes tab right? You could always play around with the order of these.

ElDeveloper:
Thanks for the prompt reviews y'all!

emperor/core.py
antgonza: I'm not sure if I follow, do you mean that you will be able to create N custom axes and then select to only show, up to 2 custom axes?

ElDeveloper: Kinda, you will always be able to select up to 3 custom axes, but you will be able to choose from however many are present in your coordinates. One thing I should have clarified before is that when we add a custom axis, we still preserve (unless otherwise specified) at least 5 dimensions of the PCoA matrix, so that means that irregardless of the number of custom axes, users can always switch to those other axes with the `Axes` tab.

antgonza: K, thanks.

emperor/support_files/js/sceneplotview3d.js
antgonza: Is there any other way around? I think it's not really intuitive and more when you print the plot and you don't know which axis is which.

ElDeveloper:
In the produced plot the percentages are removed when this value is less
than zero, and instead of showing "PC X" it displays the axis name.  The
labels are provided by the axes name attribute of the DecompositionModel
class.

antgonza:
K, thanks.

ElDeveloper:
Thanks everyone!!

Also turns out GitHub does not thread messages when you reply to them from your e-mail, oh well.

PR #541, opened 2016-10-06
MAINT: Fix examples

ElDeveloper:
Since we added the SVGRenderer as a dependency the main config broke,
this should fix it.


mortonjt:
👍 once the tests pass

mortonjt:
Although, I'm noticing that there is the big Emperor logo in the middle of the screen when I open this up in the notebook ...  Should we raise an issue for this?

mortonjt:
screen shot 2016-10-05 at 6 36 41 pm

ElDeveloper:
Thanks for the screenshot, that's super strange. I had never seen that, can you try loading the same plot in another tab?

mortonjt:
Found the problem on my end - it was pulling from an old version.

👍 @antgonza want to double check?

ElDeveloper:
thanks 👏

PR #543, opened 2016-10-13
New api biplot

mortonjt:
Here's some code for implementing biplots.

Marked a TODO: flag around parts that I don't how to implement.

Not sure what is going with the indentation - sorry!


ElDeveloper:
@mortonjt, at long last I've been able to look over it, ⏳! Thanks for your patience.

I think this is a great start! Have you had a chance to test try this out? If I recall correctly, I believe you had mentioned that it would be fine to display the feature data as spheres, is this still correct? If so, I think this should just work 😱! 

As for the tabs, we will need to clean these up. In general, in JavaScript, we want to have everything indented by two spaces.

mortonjt:
I haven't tested this out -- there is a still big chunk missing concerning the metadata handling for the features.  And I am not comfortable enough with the codebase to implement this part.

I will not have time to work on this in the foreseeable future.  If someone else wants to take the lead on this - they are more than welcome.

Issue #420, opened 2015-09-01
Unable to deselect categories on visibility tab


ajsarmstrong:
I am trying to deselect the last few samples on the visibility tab. When I scroll down to the end and click one of the boxes the window is scrolled up ~5 samples and the box I was trying to uncheck remains checked.

ElDeveloper:
Thanks for the bug report @ajsarmstrong, I was able to reproduce this problem (others, see the attached animated gif) in Chrome but not in Safari, I'll look into this as soon as possible. For the time being, it seems that this is not a problem with Safari nor with Firefox. @ajsarmstrong if you can I would suggest using a different browser for the time being, not sure what changed recently in Chrome.

![bug-report](https://cloud.githubusercontent.com/assets/375307/9642006/9e49db36-516e-11e5-9c68-db1672a1d1e0.gif)

ajsarmstrong:
@ElDeveloper thanks for looking into this. A browser change is a pretty easy work around.

gregcaporaso:
I ran to this as well today while trying to help @pandan74. The Firefox work-around works for him, but just wanted to tag him on this so he can get notifications about fixes. Thanks for working on it!

PR #545, opened 2016-10-23
Fixes styling problems

ElDeveloper:
Fixes a problem with the tables not showing properly (see #420). It additionally fixes other problems we've had with the sidebar. If all goes well, things should look just fine in Chrome/Safari/Firefox.

-------

Note this is a pull against master, not new-api.


mortonjt:
I think this is good to merge.

ElDeveloper:
Thanks both!

PR #546, opened 2016-10-27
Miscellaneous animation fixes

ElDeveloper:
- Fix problem with speed slider.
- Change default radius for animation tubes.
- Cherry pick changes from #514 to get JavaScript tests to run.


antgonza:
👍 , BTW when are we going to have McHelper back?

ElDeveloper:
Thanks for reviewing, if this looks good would you mind merging?
McHelper will be a bit hard to resuscitate, since we moved the system
and we have a different default branch. I just tried to fix the problems
right now, and am also having permissions problems, so if that's OK with
you let's leave that for later.

ElDeveloper:
Thanks @antgonza 👏 !!

PR #547, opened 2016-10-30
DOC: Add animations tutorial

ElDeveloper:
This adds a tutorial on how to create animations and a brief explanation of the concepts behind them.


doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
+.. _animations:
+
+.. index:: animations
+
+How to create an animation using Emperor
```
antgonza: Is this a question or will it be better to say: "Creating an animation using Emperor"?

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In this tutorial we describe how to create a principal coordinates analysis
+(PCoA) plot, and display animated traces of the samples sorted by a metadata
+category.
+
+To do this, we need to have two metadata categories, a *gradient* category, and
+a *trajectory* category. The *gradient* category determines the order in which
```
antgonza: Have we always call it trajectory? IMOO it sounds confusing cause the trajectory is given by the gradient, right? Perhaps grouping is a better name but OK if this is how we always have call it.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+samples are connected together, the *trajectory* category determines how
+samples are grouped together.
+
+In most cases these metadata columns already exist as part of your sample
+information, however you may need to do some curation to make these compatible
+with Emperor. As an example, in a longitudinal study where you wish to track
+the oral microbiome changes in a cohort of 3 mice over the course of 5 weeks,
+each sample will be described by the following columns:
+
+* ``cage_number``: the cage where each mice was housed, more than one mice could
```
antgonza: AFAIK, you don't need this for the animation, right? But perhaps I need to read further down.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+  have resided in the same cage.
+
+* ``age_in_years``: the age of each mice in years.
+
```
antgonza: I guess we are planning to describe 2 trajectories examples, not only one, right?

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+* ``week``: the number of the week in this experiment.
+
+* ``sex``: the sex of each mice.
```
antgonza: Is this category really need it by the animation? Perhaps adding some text after the description saying why we want it will be useful.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+
+* ``mice_identifier``: where each mice is assigned a unique identifier.
+
+In the example above, we can use the ``week`` column as our *gradient* category,
```
antgonza: Which example? I think we are still in the example ... 😉

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+so as long as all the values are numerical. To be more precise, a column where
+values were indicated as ``pre-treatment, first, second, third and last`` would
```
antgonza: This is super important and perhaps it should be part of the description of each category and why we want them.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,150 @@
...
+   unmodified ordination coloring samples by mice. On the center, the same
+   odination with a label for each sample, corresponding to the week where this
+   sample was collected. On the right, samples connected by a line, where
+   the order is determined by the collection time.
+
+
+In the next section we will go through an example using published data from
+`Weingarden et al. 2015 <https://www.ncbi.nlm.nih.gov/pubmed/25825673>`_.
+
+Data
```
antgonza: Why do we need 2 examples that are different mice/human? If you want to add them both suggest splitting them clearly.

ElDeveloper: We don't really need both, but I think that if you are unfamiliar with the topic, it may be of more help to look at a small synthetic example with just a handful of data points. As per your other comments, I've made the distinction between these two more explicit, hopefully this will help.

doc/source/tutorials/animations.rst
ElDeveloper: I've added this two paragraphs after the description of these columns.

doc/source/tutorials/animations.rst
ElDeveloper: Great idea!

ElDeveloper:
@antgonza thanks for the review, I think most of your comments have been addressed, let me know if this looks better now.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,167 @@
+.. _animations:
+
+.. index:: animations
+
+Creating an animation using Emperor
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In this tutorial we describe how to create a principal coordinates analysis
+(PCoA) plot, and display animated traces of the samples sorted by a metadata
+category. To do this, we will describe a `Synthetic Example` (explaining
```
antgonza: To do this -> Here | For this purpose,  ... Just cause that's how the next paragraph starts.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,167 @@
...
+not explicitly needed to create an animation, but can be used to change the
+color, visibility and size of the samples.
+
+The following figure shows what we expect to observe when we press the play
+button.
+
+.. figure:: trajectories.png
+   :alt: Cartoon representation of the example above.
+
+   Cartoon representation of the example we see above. On the left, the
```
antgonza: we see above -> of the synthetic example

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,167 @@
...
+   unmodified ordination coloring samples by mice. On the center, the same
+   odination with a label for each sample, corresponding to the week where this
```
antgonza: odination -> ordination

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,167 @@
...
+
+In the next section we will go through an example using published data from
+`Weingarden et al. 2015 <https://www.ncbi.nlm.nih.gov/pubmed/25825673>`_.
+
+Real Example
+============
+
+----
+Data
+----
```
antgonza: Could you use these same divisions (data, processing, filtering) in the synthetic example? Also, could you  add a small description of each section before starting the examples.

doc/source/tutorials/animations.rst
```diff
@@ -0,0 +1,167 @@
...
+vaginal communities).
+
+For convenience, we combined the two datasets using `Qiita
+<https://qiita.ucsd.edu>`_. Specifically the studies we used are `study 10057
+<https://qiita.ucsd.edu/study/description/10057>`_ (FMT) and `study 1928
+<https://qiita.ucsd.edu/study/description/1928>`_ (HMP). Remember you need to
+be logged in to access the studies.
+
+The files needed for this tutorial can be downloaded from this `link
+<http://emperor.microbio.me/animations-tutorial.zip>`_, download and unzip in
```
antgonza: I think the download and unzip is obvious but if you want to leave it change to: '. Download and unzip in'

doc/source/tutorials/animations.rst
ElDeveloper: I've now separated the sections of the two examples. The filtering section is really a "you should know" thing, not really part of the second example. I've changed the headers to make this distinction clearer.

ElDeveloper:
I've now made these changes, thanks!

antgonza:
Thanks!! Looks great. One more request, which I just thought of: how to deal with missing data? In specific, what happens if one mouse has only -1, 3?

ElDeveloper:
@antgonza good point, I had already considered this in the red samples, but did not highlight this in the text, this has now been mentioned in the text (right after the figure).

ElDeveloper:
Thanks @antgonza!

PR #549, opened 2016-11-02
REL: Release version 0.9.60

ElDeveloper:
I Incorrectly assumed 0.9.6 was greater than 0.9.51, so I have to correct this to be 0.9.60, oh the humanity.


ElDeveloper:
😞

PR #553, opened 2016-11-11
PERF: Lazily load large studies to improve performance

ElDeveloper:
When matrices with more than 1000 samples are loaded in emperor, none of the tabs are "active" (meaning that there is no table of colors, shapes, scales, etc) until a metadata category is selected for each of the controllers. After a metadata category is selected, the grid with the colors, shapes or scales is displayed, and the program should continue to work as usual.

Making these changes allows us to load the EMP (~30K samples) in about 7-10 seconds, the memory consumption is greatly reduced and the overall experience is smoother.

In making these changes, I had to make a few additional changes:

- Sort metadata headers alphabetically.
- Make the `setEnabled` method disable the widgets in the current view controller.
- Allow for `setMetadataField` to take `null` as a parameter, this in turn resets the class' attribute to its default state (for example in `ColorViewController` all markers are changed back to red, in `ScaleViewController` all markers are changed back to have a scale of 1, etc).
- Allow fromJSON and toJSON to deal with `null` values.
- Add tests for all these changes 📓.

### DIFF
--- a/emperor/support_files/js/view-controller.js
+++ b/emperor/support_files/js/view-controller.js
@@ -144,6 +144,8 @@ define([
 
     // there's a few attributes we can only set on "ready" so list them up here
     $(function() {
+      var placeholder = 'Select a ' + scope.title + ' Category';
+
       // setup the slick grid
       scope._buildGrid(options);
 
@@ -151,12 +153,11 @@ define([
 
       // setup chosen
       scope.$select.chosen({width: '100%', search_contains: true,
-                            include_group_label_in_selected: true});
+                            include_group_label_in_selected: true,
+                            placeholder_text_single: placeholder});
 
       // only subclasses will provide this callback
       if (options.categorySelectionCallback !== undefined) {
-
-        scope.$select.chosen().change(options.categorySelectionCallback);
 
         // If the number of samples is not too large, fire a callback using the
         // selector to initialize the data grid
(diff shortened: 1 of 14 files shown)


emperor/support_files/js/view-controller.js
```diff
@@ -144,23 +144,39 @@ define([
...
-                            include_group_label_in_selected: true});
+                            include_group_label_in_selected: true,
+                            placeholder_text_single: placeholder});
 
       // only subclasses will provide this callback
       if (options.categorySelectionCallback !== undefined) {
+
+        // If the number of samples is not too large, fire a callback using the
+        // selector to initialize the data grid
+        if (dm.ids.length < 1000) {
```
antgonza: I think this is an excellent idea but just wondering if we should add a message about it (different behavior) or simply always "start  grayed out", what do you think?

ElDeveloper: Thanks! Where do you think the message would fit nicely without becoming redundant? For now, a message is shown on the metadata dropdown widget "Select a Color Category" in the Color tab, or "Select a Scale Category" in the Scale tab:

screen shot 2016-11-11 at 12 21 19 pm

This is done using the tab's "name".

ElDeveloper:
@antgonza, changes should be in place. No metadata will be "auto-selected". This will overall improve the performance for everyone, not only for large studies.

antgonza:
Great, thanks!!

ElDeveloper:
Thanks for reviewing!

Issue #303, opened 2014-09-16
2D PCoA plot feature?


melisanahtar:
Could you add a feature to display the PCoA plot in two dimensions, mostly for publication purposes? Some statisticians who I've shown my 3D plots to think it's incorrect to display three dimensions on a 2D piece of paper. They're also concerned that it enables you to potentially mask data points that don't fit in nicely with your clustering by fiddling with the orientation. I've been getting around this by carefully orienting the 3D plot to look two-dimensional, but it would be great if it could be a one-click feature. Thanks for your consideration!

ozkurt:
It would be great if you can 2D plot feature. 3D is good when it is interactive, but for publication purposes we need 2Ds.
Thank you..

PR #554, opened 2016-11-12
2d plotting

wasade:
@ElDeveloper, 2D plots with a notebook example.


wasade:
Ah!!! i just noticed why fig1c is a bit more different than expected: it was only done originally with adults. PR is good for review but let me address that aspect of the example notebook before merge please.

wasade:
Okay,  added a note as to why fig1 is different. fig S4A isn't exactly the same either but close.

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+        A subplot to plot into
+    field : str
+        A metadata field in the mapping to plot.
+    mapping : pd.DataFrame
+        The metadata about the samples.
+    ordination : skbio.OrdinationResults
+        The ordination object.
+    xaxis : integer
+        An axis to plot from the ordination.
+    yaxis : iteger
```
mortonjt: iterger -> integer

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    colormap : str
+        The colormap to use.
+    null_kw : dict
+        Additional arguments to the scatter plot for the samples in the
+        background.
+    focus_kw : dict
+        Additional arguments to the scatter plot for the samples in focus.
+    """
+    mapping = mapping.loc[ordination.samples.index]
+    field_values = mapping[~mapping[field].isnull()][field]
```
mortonjt: I didn't know about this operation.  Neat!

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    ax.scatter(x, y, **null_kw)
+
+    # plot samples of interest
+    x = x_full.loc[field_values.index].values
+    y = y_full.loc[field_values.index].values
+    ax.scatter(x, y, color=colors[order], **focus_kw)
+
+    # finalize
+    ax.set_xlim(x_full.min() - 0.01, x_full.max() + 0.01)
+    ax.set_ylim(y_full.min() - 0.01, y_full.max() + 0.01)
```
mortonjt: Would be nice to have this 0.1 buffer encoded as a parameter.

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+        A subplot to plot into
+    field : str
+        A metadata field in the mapping to plot.
+    mapping : pd.DataFrame
+        The metadata about the samples.
+    ordination : skbio.OrdinationResults
+        The ordination object.
+    xaxis : integer
+        An axis to plot from the ordination.
+    yaxis : iteger
```
mortonjt: iteger -> integer

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    for value in field_values.unique():
+        value_index = field_values[field_values == value].index
+
+        x = x_full.loc[value_index].values
+        y = y_full.loc[value_index].values
+        ax.scatter(x, y, color=colors.get(value), **focus_kw)
+
+    # finalize
+    ax.set_xlim(x_full.min() - 0.01, x_full.max() + 0.01)
+    ax.set_ylim(y_full.min() - 0.01, y_full.max() + 0.01)
```
mortonjt: same here about the buffer

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    ----------
+    ordination : OrdinationResults
+        An OrdinationResults object
+    offset : int
+        A starting offset
+    mapping : DataFrame
+        The covariate information for the samples.
+    field : str
+        The field within the mapping file to use for coloring.
+    colors : dict, str
```
mortonjt: dict, str -> dict or str

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    PC1v2 PC1v3 PC1v4 BLANK
+    PC2v3 PC2v4 PC2v5 Legend
+    PC3v4 PC3v5 PC3v6 BLANK
+
+    Parameters
+    ----------
+    ordination : OrdinationResults
+        An OrdinationResults object
+    offset : int
+        A starting offset
```
mortonjt: It is not exactly clear what this offset is referring to.  Is it the spacing in between subplots?

mortonjt:
This is a really good first pass @wasade !

When it comes to developing wrappers for matplotlib, I would try to use [seaborn](https://github.com/mwaskom/seaborn) as a reference.

Specifically

1. There is no reason why the plots have to be only 3x3.  If you could take advantage of [FacetGrid](http://seaborn.pydata.org/generated/seaborn.FacetGrid.html), you could either only push in code for a single plot, and use the [map](http://seaborn.pydata.org/tutorial/axis_grids.html
) functionality.  The other alternative is to just wrap and return a FacetGrid.  Either strategy could help simplify the code, and reduce the amount of hard-coding

2. Seaborn actually has really nice strategy for testing plots.  It assumes that matplotlib is always correct, and only tests the internals within matplotlib.  You can see this [response](http://stackoverflow.com/a/27950953/1167475) on SO.  This is the strategy that @jkanbar took when developing the biplot code [here](https://github.com/mortonjt/canvas/blob/master/canvas/tests/test_biplot.py).

3. It is definitely preferable to return an axes object or a FacetGrid than a figure object.  This is because this assumes that only one figure will be returned from this method.  There are scenarios where this method is just another subfigure in a larger figure.  And changing the returns object would simplify this.

emperor/flat_plots.py
wasade: done

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    mapping = mapping.loc[ordination.samples.index]
+    field_values = mapping[~mapping[field].isnull()][field]
+
+    nonfield_samples = set(mapping.index) - set(field_values.index)
+
+    # discretize our color gradient, and determine which bin (i.e., color)
+    # a given value is associated with
+    bins = np.arange(field_values.min(),
+                     field_values.max(),
+                     field_values.max() / 256)
```
wasade: yes, good catch, thanks!

emperor/flat_plots.py
wasade: done

emperor/flat_plots.py
wasade: done

emperor/flat_plots.py
wasade: done

emperor/flat_plots.py
wasade: done

emperor/flat_plots.py
```diff
@@ -0,0 +1,337 @@
...
+    PC3v4 PC3v5 PC3v6 BLANK
+
+    Parameters
+    ----------
+    ordination : OrdinationResults
+        An OrdinationResults object
+    offset : int
+        A starting offset
+    mapping : DataFrame
+        The covariate information for the samples.
```
wasade: reusing the doc from `core.py`

wasade:
@mortonjt, there are many libraries that can be used to do the same thing and I do not feel compelled to redo this using a different library. 

There is no requirement that the plots are 3x3, that is simply the driver function which I need right now and the underlying code used per set of axes is can be reused. In addition, please note that the 3x3 method is already returning the axes in addition to the figure. 

The current strategy of exercising the code was already discussed with @ElDeveloper.

emperor/flat_plots.py
```diff
@@ -0,0 +1,346 @@
...
+        If a dict, it is expected to provide a mapping of category value to
+        a color; entries not in the mapping will be considered "null". If a
+        str, the value is interpreted as the name of a color map and assumes
+        a gradient is to be painted.
+    null_kw : dict, optional
+        Parameters to use for plotting and painting null or unknown values.
+    focus_kw : dict, optional
+        Parameters to use for plotting and painting values which are not
+        null and in the field of interest.
+    """
```
mortonjt: It would be nice to add a note that there is an intersection operation occurring - that only the samples contained in both the mapping file and the ordination results will be considered.

mortonjt:
Hi @wasade - I have no hard feelings about using seaborn or not.  

Concerning the figure development, this is still not ideal for embedding into subplots, since an entire new figure is being built.  If you wanted to actually embed the subplots into an existing figure, you'd have to extract out all of the axes objects from the new figure to embed in the existing figure.  But I'm not entirely sure about the scope of this PR - @ElDeveloper would need to weigh in on this.

About the tests - I do think that there should be a few more checks.  Right now, only failure checks are being made.  Specifically, it would be nice to see the following tests

1. Test to make sure that there are 9 main subplots
2. Test to make sure that the points in one of the subplots is correct (you can do this by extracting the matplotlib objects and testing for the contents).  This also includes testing the axes to make sure that the labels are correct
3. Test to make sure that the contents of the legend are actually correct for both discrete and continuous

This tests are more to ensure that silent bugs aren't being introduced in the future - which would be extremely hard to catch without some basic testing functionality.

wasade:
This is failing due to a failure to connect to the display despite the `use('Agg')`. Any ideas?

@mortonjt, the underlying methods for generating the plots simply take an axes object and a separate PR can put in another driver method. The `plot_3x3` method is how I would like it to be, and I do not see a usecase in embedding such a large figure somewhere else.

I'm still not convinced of the necessity of additional testing of the content of figures as there is already an integration test via the notebook. Can someone else please weigh in? Alternatively, I can close this PR out and push it and the code to the EMP repository as that's what this code was originally developed for, although there has been a long running request to get 2D plots into EMPeror (#303).

antgonza:
display error: not sure but I think this is how we solved it in Qiita: https://github.com/qiita-spots/qtp-target-gene/blob/master/.travis.yml#L6

wasade:
Trying, thanks!

On Mon, Nov 14, 2016 at 11:58 AM, Antonio Gonzalez <notifications@github.com

antgonza:
Should have mentioned that is that line and the next one. Suggest letting the test run and if fail, adding the next line.

wasade:
Failures now from numpy version...? What's the motivation to test with older versions of numpy btw?

ElDeveloper:
The motivation is to allow for compatibility with whatever package that
relies on older versions, historically this was a consequence of QIIME,
and the fact that it depended on a very specific version of NumPy.

antgonza:
I guess that means that we can update?

ElDeveloper:
I think so, so as long as we are compatible with skbio, qiime2 and qiita (though IIRC qiita doesn't really care), I think it will be fine to do whatever.

wasade:
Okay, thanks. Second, it looks like the different version of numpy coerces a different version of matplotlib (1.4.3) in which "viridis" is not a valid color palette. I can switch the palette used in the tests, but that raises the question of whether matplotlib 1.4.3 should be allowed or disallowed?

emperor/flat_plots.py
```diff
@@ -0,0 +1,346 @@
+from matplotlib import use
+use('Agg')  # noqa
+import matplotlib.pyplot as plt
+import numpy as np
+import seaborn as sns
+
+
+def plot_3x3(ordination, offset, mapping, field, colors, null_kw=None,
```
ElDeveloper: Can we make this such that an arbitrary number of dimensions is plotted? i.e. pc1 vs pc2 alone? For example, an interface like this would be more flexible:

```python

def daniels_great_plotting_function(ordination, mapping, colors, field, cols, offset, null_kw=None, focus_kw=None):
     # ...
```

The main differences are the order, and the addition of a `cols` argument that lets you specify which dimensions are to be plotted, for example if you were to pass `['pc1', 'pc2']` as the value of `cols`, then you would only get as a result  a single scatter plot showing you pc1 vs pc2.

Additionally, it would be great if the colors could be determined from the current color list (as noted by the matplotlib API), this way there's no need to create a specific map between the categories and a color, though having the ability to do so is certainly useful. In this way, I think we want to mimic seaborn's API.

Together these changes would get rid of the `offset` argument and simplify the example in your notebook.

Lastly, have you thought about having shapes be used in these plots? That's a nice thing we could leverage from matplotlib.

ElDeveloper:
PS - The histogram for the continuous plot looks great!

emperor/flat_plots.py
```diff
@@ -0,0 +1,346 @@
+from matplotlib import use
```
wasade: was for travis but the subsequent monkeying may have obviated the need

emperor/flat_plots.py
wasade: I think this request is for a fundamentally different driver method and is in my opinion out of scope for this PR. The layout is intentionally 3x3 and ridged as it ensures consistency between multiple figure sets such that flipping through say, Preview.app, allows for seeing relationships between different sets of metadata.

The color mapping functionality is required for the current use case with the EMP. I'd be happy to consider the case of `colors=None` and let matplotlib resolve the colors. 

Shapes are supported as you can pass in figure specific parameters with `focus_kw` and `null_kw` although they are not category value specific.

wasade:
@ElDeveloper, it is possible to pass in axes objects to the private methods already. `plot_3x3` is specific to a presentation format and layout. I disagree that it makes sense to pass in an axes object.

wasade:
And thanks re: histogram :)

mortonjt:
Is it possible to have a separate PR for a single plot?  Because there are scenarios where having just a single 2D plot would be useful.

re tests - I strongly recommend putting them in.  The tests currently in place are shallow, and won't be able to catch silent bugs in future iterations.  We're talking about ~5 additional unittests, so it should be really easy to add these in.  I can also push in a PR against your branch to speed up this process.

wasade:
@mortonjt, the code for the individual 2D plots are already included, `_two_dimensional_gradient` and `_two_dimensional_discrete`. These have been renamed to be denoted as public.

@ElDeveloper, handling the case of not specifying a colors `dict` for discrete is not trivial as the determination of a category value in focus is dependent on having an associated color. Is this a blocking request?

wasade:
Tests are in and I removed the numpy versions

josenavas:
Just want to drop a comment here, I don't think it fully applies here but wanted to note. If understand correctly, this is generating a non-interactive 2d plot, right?

Note that 2d plots can be accomplished in the normal emperor 3d view by setting the camera to "orthogonal" projection (instead of "perspective"), locating the camera in the right position and disabling the camera rotations. This will give the illusion of a 2D plot although the entire scene is a 3d, and all other functionality from emperor would be available.

wasade:
@josenavas emperor doesn't scale to the size datasets I'm working with. Even on small datasets though, the amount of manual effort necessary to reproduce what is in this PR is large.

josenavas:
That makes sense! just dropping my 2 cents :)

emperor/flat_plots.py
```diff
@@ -0,0 +1,352 @@
...
+    ax : AxesSubplot
+        A subplot to plot into
+    field : str
+        A metadata field in the mapping to plot.
+    mapping : pd.DataFrame
+        The metadata about the samples.
+    ordination : skbio.OrdinationResults
+        The ordination object.
+    xaxis : integer
+        An axis to plot from the ordination.
```
mortonjt: Shouldn't this be a string?

wasade: nope

emperor/flat_plots.py
```diff
@@ -0,0 +1,352 @@
+import matplotlib.pyplot as plt
+import numpy as np
+import seaborn as sns
```
mortonjt: We're depending on seaborn?

emperor/flat_plots.py
```diff
@@ -0,0 +1,352 @@
...
+    nonfield_samples = set(mapping.index) - set(field_values.index)
+
+    # discretize our color gradient, and determine which bin (i.e., color)
+    # a given value is associated with
+    bins = np.arange(field_values.min(),
+                     field_values.max(),
+                     (field_values.max() - field_values.min()) / 256.0)
+
+    order = np.digitize(field_values, bins=bins, right=True)
+    colors = np.array(sns.color_palette(colormap, n_colors=len(bins) + 1))
```
mortonjt: Shouldn't we be using a [matplotlib colormap](http://matplotlib.org/api/cm_api.html) instead?

emperor/flat_plots.py
wasade: for the palette

emperor/flat_plots.py
wasade: if it works the same or similar

mortonjt:
Note that we'll also want to add in a test explicitly testing for the Ordination results objects - just in case that object changes in skbio

emperor/flat_plots.py
mortonjt: ok cool - then we can probably ditch that dependency

wasade:
I don't follow re: `OrdinationResults`. If there are API concerns, we can just restrict `setup.py` to `skbio >= 0.5.0, < 0.6.0`

emperor/flat_plots.py
mortonjt: hmm - that's missing quite a few use cases.

Most of the scikit-bio instances of the OrdinationResults objects are populated with string labels

https://github.com/biocore/scikit-bio/blob/master/skbio/stats/ordination/_principal_coordinate_analysis.py#L136

wasade: Needs to be both apparently. The PC files from QIIME 1.9.1 are not strings and are a `RangeIndex` while the results from `skbio.stats.ordination.pcoa` are `str`. That's annoying.

mortonjt: I think it is ok to enable either strings or ints.  You just need to update the docs to specify that either str or int can be passed in.

emperor/flat_plots.py
mortonjt: this can be done as follows
```
    cNorm = colors.Normalize(vmin=field_values.min(), vmax=field_values.max())
    scalarMap = cmx.ScalarMappable(norm=cNorm, cmap=colormap)
    colors = scalarMap.to_rgba(np.linspace(0, 1, 256))
```

mortonjt:
Hi @wasade I've done a complete review on this

1. Can you a wrapper method as follows
```python
def scatter2d(field, mapping, ordination, xaxis, yaxis,
              colormap, null_kw, focus_kw, ax=None):
    if ax is None:
        f, ax = plt.subplots()

    if isinstance(colormap, dict):
        _two_dimensional_discrete(ax, field, mapping, ordination, xaxis, yaxis,
                                  colors, null_kw, focus_kw):
    elif isinstance(colormap, str):
        _two_dimensional_gradient(ax, field, mapping, ordination, xaxis, yaxis,
                              colormap, null_kw, focus_kw):

    return ax
```
Because `_two_dimensional_discrete` and `_two_dimensional_gradient` only differ by 1 keyword

2. Can you swap out the seaborn palette with matplotlib as mentioned before?

I've looked over the tests more carefully and in fact it looks like they are complete.  I think this can be merged once the above comments have been addressed.

emperor/flat_plots.py
ElDeveloper: I have to acknowledge that I did not realize from your original demonstration (during the EMP meeting), that the resulting plots actually display up to PC6. I misunderstood this, my apologies for overlooking that. While I understand your concerns with managing space distribution (and all the cumbersome matplotlib juggling that that entails), it seems like there could still be a way to guarantee consistency, if we opt for a slightly different layout. In the image below I have drawn two cases, when someone selects PC1, PC2 & PC3 and when someone selects PC1, PC2, PC3 & PC4.

(tilt your head 🙃 - there's no 90º tilted emoji )
img_4502

What would essentially be getting drawn are the upper diagonals of the matrix of combinations of the categories. And this would open up space for a different placement of the histogram. It is true that this functionality differs a lot from what is currently implemented, so do let me know if this is not something you would be interested in implementing. Regardless, I think that's the API we want to have, providing only 3x3 plots seems rather arbitrary and inflexible.

Adding shapes based on a category will be very useful, but we should defer that to a separate iteration.

ElDeveloper:
> emperor doesn't scale to the size datasets I'm working with. Even on small datasets though, the amount of manual effort necessary to reproduce what is in this PR is large.

I want to clarify, these plots can be loaded in the GUI (see #553), the problem is exporting them in a reasonable format for publication, SVGs take a while to be generated with these many polygons and end up choking the browser. In any case having the ability to programmatically generate figures from the Python interface has been a recurrent use-case, and we can't quite support that using the GUI (without some hacking to run a headless browser), so I think the matplotlib approach is appropriate here.

emperor/flat_plots.py
wasade: @ElDeveloper, `plot_3x3` is just one possible layout, and is one which I am currently using and which fits my usecase well. The underlying methods to produce individual plots are exposed; other layout support would be great but in my opinion, it is out of scope for this PR.

ElDeveloper: @wasade I will take over making these changes to convert plot_3x3 into something like the API I described above. Having this layout indeed fits your usecase, but is too specific.

emperor/flat_plots.py
wasade: the gradients which result from this are pretty terrible. Can you verify this is equivalent? Please see the attached screenshots
screen shot 2016-11-16 at 11 45 14 am
screen shot 2016-11-16 at 11 45 20 am

rob-knight:
Is there no way to render them to a higher resolution virtual display and save screen capture? Given that many laptops can drive 4K displays seems like there should be some option?

On Nov 16, 2016, at 10:56 AM, Yoshiki V?zquez Baeza <notifications@github.com<mailto:notifications@github.com>> wrote:


emperor doesn't scale to the size datasets I'm working with. Even on small datasets though, the amount of manual effort necessary to reproduce what is in this PR is large.

I want to clarify, these plots can be loaded in the GUI (see #553<https://github.com/biocore/emperor/pull/553>), the problem is exporting them in a reasonable format for publication, SVGs take a while to be generated with these many polygons and end up choking the browser. In any case having the ability to programmatically generate figures from the Python interface has been a recurrent use-case, and we can't quite support that using the GUI (without some hacking to run a headless browser), so I think the matplotlib approach is appropriate here.

-

ElDeveloper:
@rob-knight We do have the ability to render as PNGs directly from the interface, if the size of the browser's window is big enough, then saving the PNG would be equivalent to doing what you suggest.

rob-knight:
Can you use js to expand window to bigger than real screen and export? Then reset back to orig size?

On Nov 16, 2016, at 5:16 PM, Yoshiki V?zquez Baeza <notifications@github.com<mailto:notifications@github.com>> wrote:


@rob-knight<https://github.com/rob-knight> We do have the ability to render as PNGs directly from the interface, if the size of the browser's window is big enough, then saving the PNG would be equivalent to doing what you suggest.

-

ElDeveloper:
I will give this a shot and report back!

mortonjt:
@wasade I have just made a PR against your branch removing the seaborn import.

Once that is merged in, there are only 2 more check boxes that need to be addressed!

mortonjt:
What do you guys think about closing this PR?  I think it would be more worthwhile to consider redoing this plots with Bokeh, especially considering that bokeh can handle webgl out of the box.

wasade:
Can bokeh handle the EMP?

mortonjt:
Definitely

https://groups.google.com/a/continuum.io/forum/#!topic/bokeh/vpjbJB67Vjs
http://ryanhafen.com/blog/plot-lots-of-data

Bokeh can handle much larger datasets then what we are dealing with.

antgonza:
Note that in general the problem is not the number of spheres/points but their combination with metadata ... I think is worth investigating how feasible is to use bokeh and how will it interact with our current code before closing or making a larger effort.

wasade:
Only if we require webgl for them, which is probably fine. And I agree with @antgonza on the metadata.

What I'm worried about is that we took something which was already useful, and instead of getting it merged, it got forced into limbo where no one can use it. I'm fine with whatever path forward makes these types of plots accessible to users either with the current form of this PR or with some other underlying plotting library.

ElDeveloper:
I think matplotlib can handle the size of our data, I remember @wasade generating these plots with his laptop in a matter of seconds. Bokeh is an excellent alternative, but I don't see any pressing need to move to it, of course if this was moved to bokeh and it integrated the modifications to the API suggested above, I think that would also be fine.

josenavas:
Sorry - jumping in late on the discussion. I thought 2D plots could be accomplished with a static orthogonal camera? (The current camera that we have is a perspective camera)

ElDeveloper:
@josenavas, yes that can indeed be achieved and I am still planning on doing that, though I think there's still a situation where you need 2d plots that you programmatically generate and modify. Something like the current approach will make sense.

PR #556, opened 2016-11-14
Update documentation for 1.0.0 beta

ElDeveloper:
In this PR I remove all the documentation that was created with version 0.9.x in mind, and add docs specific to the new 1.0.0 version. For now, I've left the `Description` document in an incomplete state as the interface itself is not finalized yet, so I think we will be able to fill this in as we make more progress on this front.


README.md
```diff
@@ -13,26 +13,10 @@ Before contributing code to Emperor, please familiarize yourself with the [contr
 
 ## Usage examples
 
-The main interface to create Emperor visualizations is the `make_emperor.py` script, inputing a mapping file and a PCoA data file, will generate an Emperor graphical user interface to analyze and visualize your data.
+We recommend using Emperor through its Python interface. The following, are two
+examples of how to do this:
 
-If you have a QIIME compliant mapping file and a PCoA file, try the following command from a terminal session:
+- [Visualizing data produced with scikit-bio](http://nbviewer.jupyter.org/github/biocore/emperor/blob/new-api/examples/keyboard.ipynb).
```
antgonza: ❓ is that really the recommended way? Are we planning to have something similar to make_emperor.py? I like the idea of a script cause if I have a single pc/map-file and want to plot I don't need to have a full notebook.

doc/source/examples.rst
```diff
@@ -0,0 +1,108 @@
+.. _examples:
+
+.. index:: examples
+
+Examples
+^^^^^^^^
+
+Emperor provides a Python API (:ref:`emperor-python-docs`) intended to be used
```
antgonza: Same question than before, what are the benefits vs. a CLI?

antgonza: For example, what will be the shortest path to have the new emperor in Qiita?

doc/source/examples.rst
```diff
@@ -0,0 +1,108 @@
...
+examples above that's why we set ``remote=True``).  Remember, that in order to
+do this, you will need to have access to an internet connection.  Alternatively
+when there's no internet connection available, you can use the `nbinstall`
+function, which will make the resources available from within your running
+notebook server.
+
+Outside the Jupyter Notebook
+----------------------------
+
+In certain circumstances it is more convenient to create plots independent of
```
antgonza: IMOO, not ideal.

doc/source/examples.rst
ElDeveloper: The shortest path, I think, will be to use QIIME2 and its CLI. Alternatively, if we wanted, we could use the Python interface to generate the plots from within the Qiita plugin.

mortonjt:
:+1: on leveraging the Q2 api

ElDeveloper:
Thanks!

PR #557, opened 2016-11-17
BUG: Fix problem with gradients

ElDeveloper:
Gradients would fail to display when the "show continuous" button was pressed. The problem was manifested when the data was split into numeric and non-numeric data (missing data), some values were incorrectly cast to numbers, and if a value was cast from `'1.0'` to `1`, this object would later not be found anywhere and an error would be thrown. I changed the function to return unmodified objects, and perform the casts to numbers only when needed (I also added a few extra tests).

Thanks to @cuttlefishh and @amnona for reporting this problem.


mortonjt:
Umm, I'm still seeing this problem

image

mortonjt:
I approve this pull request

mortonjt:
@antgonza @wasade can you merge if you like this?  I think this resolve the issues that @amnona and @cuttlefish saw

PR #558, opened 2016-11-18
ENH: Checkboxes no longer require extra click

ElDeveloper:
Previously, to hide/show sets of samples, you would need to click on the
visibility grid to "confirm" the change, this has been fixed to
auto-confirm. Since we were using the original minified editor file,
I've modified the code in-place, but we can re-add the un-minified
version if others think it's appropriate (no strong feelings).


antgonza:
Just to confirm, this means that we have a modified version of slick.editors.min.js that could be not updated during the next upgrade, right? Is it possible to change that parameter during the configuration of the grids? If not, this is OK to merge.

ElDeveloper:
> we have a modified version of slick.editors.min.js

You are right, this is correct.

> Is it possible to change that parameter during the configuration of the grids?

No, it is not possible, the specific change is modifying a method to alter its logic.

We could keep an un-minified version of this file.

ElDeveloper:
This sounds good, I've added a few lines at the top of the minified file, but they are not showing because GitHub hides the changes of "long files", can you check if those seem fine to you?

antgonza:
Thanks!

ElDeveloper:
Thanks both!

PR #561, opened 2017-02-23
ENH: Add context menu item to recenter the camera

ElDeveloper:
This adds a new option that allows users to go back to the initial
camera position and rotation.


antgonza:
Do we really need context-menu-icons.* ?

ElDeveloper:
Unfortunately yes we do, different browsers use different resources to
display the icons in the context menu.

PR #569, opened 2017-03-29
ENH: Save mapping file and coordinates as JSON

ElDeveloper:
The data is now written using JINJA's tojson filter, this makes the
translation a bit more natural, specially for numbers in scientific
notation.

You may need to update jinja or you might get an error.


ElDeveloper:
This should be ready for review, the tests are passing ✅

ElDeveloper:
This should be ready to be merged, can someone have a look?

emperor/support_files/templates/logic-template.html
```diff
@@ -97,21 +97,29 @@
...
 
-  var ids = {{ coords_ids }};
-  var coords = {{ coords }};
-  var pct_var = {{ pct_var | map('float') | list }};
-  var md_headers = {{ md_headers}};
-  var metadata = {{  metadata }};
-  var axesNames = {{ axes_names }};
+  var data = {
+    // coordinates information
+    'sample_ids': {{ coords_ids | tojson }},
```
wasade: I'm not familiar with the `| tojson`, is that like a bash pipe in jinja?

ElDeveloper: Yeah, they call them filters in the JINJA world http://jinja.pocoo.org/docs/2.9/templates/#filters

ElDeveloper:
Thanks @wasade!

Issue #542, opened 2016-10-10
Blurry axis labels


mortonjt:
Not sure what is going on here.  Any ideas?

<img width="552" alt="screen shot 2016-10-09 at 6 46 57 pm" src="https://cloud.githubusercontent.com/assets/4184797/19225405/e9436e20-8e50-11e6-8720-2ce4b8142082.png">

Also getting the following error message in the javascript console

```
actions jupyter-notebook:find-and-replace does not exist, still binding it in case it will be defined later...
http://localhost:8888/nbextensions/widgets/notebook/js/extension.js?v=20161009122930 Failed to load resource: the server responded with a status of 404 (Not Found)
main.min.js:33970 Widgets are not available.  Please install widgetsnbextension or ipywidgets 4.0
main.min.js:23424 Session: kernel_created (cc14d8c3-3063-41f0-ab44-b863a43fee66)
main.min.js:22710 Starting WebSockets: ws://localhost:8888/api/kernels/72cd5d9d-a1f5-4c67-b87a-7905972173b8
main.min.js:22357 Kernel: kernel_connected (72cd5d9d-a1f5-4c67-b87a-7905972173b8)
main.min.js:22357 Kernel: kernel_ready (72cd5d9d-a1f5-4c67-b87a-7905972173b8)
ordination.ipynb:1 Font from origin 'null' has been blocked from loading by Cross-Origin Resource Sharing policy: Invalid response. Origin 'http://localhost:8888' is therefore not allowed access.
2cdn.rawgit.com/biocore/emperor/new-api/emperor/support_files/vendor/js/three.min.js?v=20161009122930:578 THREE.WebGLRenderer 73
```

mortonjt:
It is important to note that I am building a custom OrdinationResults object, so that may have something to do with this (the axes are 0, 1, 2 respectively)

ElDeveloper:
Whoa, I had never seen that. That's interesting. Seems like a problem with loading resources remotely?  I'm not sure, can you try using `remote=False` and see if the problem persists? Also just in case, make sure you clean the output of the notebook, save, reload the page and run `nbinstall`.

PR #570, opened 2017-03-30
Space normalization

ElDeveloper:
When dealing with arbitrary coordinate matrices in WebGL, it can be challenging to properly set visual defaults that look good for most cases, and while we had found an *ok* solution, there were still situations where some problems would arise for example see #542. Under different conditions other artifacts [would arise](https://github.com/biocore/emperor/pull/532#issuecomment-252971457).

By scaling the matrices before we display them in WebGL, these problems disappear. For an example (from @wasade):

### before
screen shot 2017-03-29 at 11 49 37 pm

### after
screen shot 2017-03-29 at 11 49 52 pm


ElDeveloper:
Thanks @antgonza!

PR #571, opened 2017-03-30
Add option to create a large screenshot 📸

ElDeveloper:
Previously we only generated SVG files and PNG files, both of which adopted the resolution of the current window. This PR adds the option to create a **large** resolution version of the plot and save it to a PNG file. This solution was suggested by @rob-knight, who [pointed out](https://github.com/biocore/emperor/pull/554#issuecomment-261124847) this as an acceptable alternative to large SVG files.

Currently we scale the resolution to be **5** times bigger, and the prints look great.

Just one open question, should we still print non-high-resolution versions of the plot?


ElDeveloper:
Maybe, but I am not sure what machines will be able to produce these images and which won't. See the comparison:

# low res
emperor-low-res

# high res
emperor-high-res


As for size, it's much larger:

```
yoshikivazquezbaeza:Desktop$ du -sh *.png
1.4M	emperor-high-res.png
100K	emperor-low-res.png
```

ElDeveloper:
Thanks!

PR #574, opened 2017-04-04
BUG: Fix typo in canvas-toBlob

ElDeveloper:
Oddly enough this bug only showed in Safari but not in Chrome, I just updated the notebooks to work again.

This will need another release 🤕


ElDeveloper:
If anyone is wondering why I merged this, is because I found a critical bug (plots wouldn't load as there was a typo in the `require.js` statements), that is fixed by this PR. I've also updated the notebooks. The reason why this was not caught before was because the unit tests have a different require.js setup file than the actual code, and that this was not a problem in Chrome (who knew).

PR #575, opened 2017-04-04
REL: Bump strings for beta 7

ElDeveloper:
Merge after #574


ElDeveloper:
Again, if someone is wondering about why I merged this, I need to make a quick release to prevent anyone installs a version of emperor with the bug fixed in #574.

PR #578, opened 2017-04-04
BUG: Resolve stable URLs depending on package version

ElDeveloper:
Deals with a problem where plots created with the `remote=True` argument would lead to the plots failing if new dependencies were added.

The base URL is now resolved depending on the version string. I had to change the tests because we can no longer test `remote=True` explicitly as the version string will change dependent of the version.

@wasade, this would deal with the need to update the figures in the deblur-manuscript repo.

------------

#### During code review this point came up

- [x] When `remote=True` is used, the code should raise a warning alerting the user of the possible consequences of these actions.

### DIFF
--- a/emperor/__init__.py
+++ b/emperor/__init__.py
@@ -35,7 +35,7 @@ Functions
 # ----------------------------------------------------------------------------
 
 import pkg_resources
-__version__ = pkg_resources.get_distribution('emperor').version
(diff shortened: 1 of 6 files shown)


ElDeveloper:
OK, this is finally passing the tests, should be ready for review and merge 🎉 🍳

ElDeveloper:
Thanks @antgonza!

PR #579, opened 2017-04-06
ENH: Allow users to set custom visualization settings from the Python API

ElDeveloper:
When working in the Jupyter notebook, users can now set visibility, coloring, scaling and shapes from the Python API. This is perhaps better explained by the following animated gif:

custom-settings-3

The above demonstration is intended to show how users can call `emp.color_by('host_subject_id')` and the result will translate in the displayed plot coloring *by default* using the `host_subject_id` category. It also shows how you can put many of these calls together, for example by changing the coloring category (to `age`) and then changing the visibility (to hide one of the subjects - `M9`).

Open points
=========

- [x] Needs to add control for axes controller.
- [x] There are references from the Python documentation to the JavaScript documentation, how pertinent is this?
- [x] In order to load the user-defined-settings right now we are waiting 1 second after the UI is loaded and then we trigger a *refresh*, this is fragile and could be better.
- [x] Individual values for the different methods do not check the validity of the data i.e. the `color_by` method doesn't verify that '#fexxxx' is a valid CSS color, so this might result in a failure when loading the plot, it however, checks that the color itself is represented as a string.
- [x] All these methods return the object itself, which is not a very common thing to do in Python, but I think makes the chaining and plotting more natural, thoughts?
- [x] Probably need to add more tests that ensure multiple settings modifications work well.
- [x] Save/load settings (view, colors, etc) via commands (serialize and deserialize the `settings` property into a JSON object).

##### Visibility-specific

- [x] Add a negate parameter to `visibility_by`.
- [x] Visibilities would make more sense as a list

##### New feature

- [x] ~~Add a strict mode to these methods~~. This doesn't make a lot of sense for some methods.

##### Settings related

- [x] Have a global reset.
- [x] ~~Use None to clean up a category~~ this was ultimately ignored as the interface became a bit confusing.
- [x] Include a `set_settings` method.
- [x] Be able to load a set of settings altogether

##### Other

- [x] Colormaps should be listed in a table.

-------------

Please wait to merge after Code Review on Monday April 10, 2017. 🏁


antgonza:
The code looks good and the list seems reasonable. I would add, save/load settings (view, colors, etc) via commands.

ElDeveloper:
@antgonza, that's a great idea!

lkursell:
I know this has been mentioned before, but I worry this makes it suspiciously easy for a user to calculate beta diversity once on the full data set, and then drop the visibility of samples until a nice separation is obtained. I'm not saying we have to be guardians for others being rigorous, but having this same filter function _before_ beta diversity is run and graphed would be scientifically "safer".

antgonza:
@lkursell, what about a warning? Not a real solution but I don't think there is a real solution for that problem ... but education.

lkursell:
@antgonza @ElDeveloper I think a warning or watermark with a statement like "This PCoA plot does not accurately reflect true distances between these samples and is not appropriate for publications" would go a long way. @johnchase has advocated for this strongly in the past, and I agree.

rob-knight:
+1

I can just imagine the cherry-picking that this makes easy, sort of like that correlation demo throwing away points I used to do for class. So if any samples that contributed to the PCoA are hidden, the watermark should be added to the plot.

On Apr 6, 2017, at 9:50 AM, Luke Ursell <notifications@github.com<mailto:notifications@github.com>> wrote:


@antgonza<https://github.com/antgonza> @ElDeveloper<https://github.com/ElDeveloper> I think a warning or watermark with a statement like "This PCoA plot does not accurately reflect true distances between these samples and is not appropriate for publications" would go a long way. @johnchase<https://github.com/johnchase> has advocated for this strongly in the past, and I agree.

ElDeveloper:
This is true, maybe the way to go is by raising a warning whenever the `visibility_by` method is called, or perhaps wait for it to be called 3-4 times? Or alternatively, I can add an overlaid message on the plot warning people that samples have been hidden, and that this plot may be prone to misinterpretations.

Alternatively we could have the "I am a dummy" mode, like mysql, though this is too snarky (I think):
http://sql-info.de/mysql/notes/I-am-a-dummy.html

wasade:
I like the overlay on by default. We could optionally make it so it could be disabled which would be nice for presentations and movies, but at the same time, being able to disable it could lead to abuse so it would be good to make it shamefully possible to disable if that route is taken.

johnchase:
I think the overlaid message is the way to go. The problem I could see with the warning is that it would be coupled to the plot only at the time it is called. If someone decided to share a plot that information would be lost.

 I do not like the idea of a "dummy mode" though it is interesting...

ElDeveloper:
This should be ready for review/merge.

ElDeveloper:
Friendly ping, this should be ready for review and merge, most of the changes are either documentation or tests. 🎆

ElDeveloper:
After talking to @antgonza, this pull request is almost complete, the last piece missing concerns the warning that users should see when they fiddle around with the visibility tab. There's a few options:

1) Showing a warning from the Python interface whenever the `visibility_by` method is called.
2) Show a [modal dialog](https://jqueryui.com/dialog/#modal-message) in the GUI **every single time** the visibility of a set of samples is modified.
3) Show a [modal dialog](https://jqueryui.com/dialog/#modal-message) in the GUI **only the first time** the visibility of a set of samples is modified.
4) Show a watermark on the plot anytime the visibility of a set of samples is modified. If implementing this option, we will need a way to disable the watermark for use-cases when hiding samples is not done maliciously, but we think that an appropriate requirement is that the JavaScript code be modified directly (perhaps a harsh but fair requirement).

Would love to hear any feedback @wasade @johnchase @lkursell @rob-knight

johnchase:
> we will need a way to disable the watermark for use-cases when hiding samples is not done maliciously

@ElDeveloper I don't think it's ever done maliciously, but what is the use case where it would be appropriate to display a subset of PCoA coordinates other than data exploration? 

I think we discussed this in the past and I apologize I don't remember the outcome, but would the appropriate way to remove the watermark be to recalculate principal coordinates on only the subset of samples to be displayed?

ElDeveloper:
Fair enough, malicious may have been an overstatement. I think it is OK to hide and show groups of samples **so as long as the complete plot is shown for a reference**. For example see Figure 4 of [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3787266/), or in animations to highlight a group, and then show the context back again.

> I think we discussed this in the past and I apologize I don't remember the outcome, but would the appropriate way to remove the watermark be to recalculate principal coordinates on only the subset of samples to be displayed?

That **is** the appropriate way, I agree. However, this would require for the UI to be linked to a server and have knowledge of the distance matrix that generates a plot, so we can't accommodate this unless we require that Jupyter is **always** running in the backend.

rob-knight:
Usually you want to turn them on and off interactively to highlight different areas in exploratory analysis.

On May 2, 2017, at 1:04 PM, John Chase <notifications@github.com<mailto:notifications@github.com>> wrote:


we will need a way to disable the watermark for use-cases when hiding samples is not done maliciously

@ElDeveloper<https://github.com/ElDeveloper> I don't think it's ever done maliciously, but what is the use case where it would be appropriate to display a subset of PCoA coordinates other than data exploration?

I think we discussed this in the past and I apologize I don't remember the outcome, but would the appropriate way to remove the watermark be to recalculate principal coordinates on only the subset of samples to be displayed?

-

johnchase:
@ElDeveloper That makes sense. I still feel it is odd to publish a plot (even if it is only to highlight specific samples) that doesn't represent the the PC values of the samples shown, however I understand the use case. To some extent we are using to emperor to inform the limitations of principal coordinates, which is not necessarily the responsibility of emperor. 

Given that my vote would be for option number 4 - watermark with the ability to turn it off.

In terms of recalculating principal coordinates I was referring to the user being responsible for recreating the plots with the subset of samples. It would be great if emperor could be aware of the distance matrix and and calculate it on the fly but I realize there is not simple solution for that.

ElDeveloper:
This seems to be a bit more complicated but I've opened #589 to keep track of this.

PR #582, opened 2017-04-25
Update THREE.js to r83

ElDeveloper:
Updates THREE.js and all plugins. Note that some of the require.js configs had to change due to some backwards incompatible changes in THREE, I opted for using the same strategy as \`ili as these changes are intended to allow the integration with `ili.

Note that we will want to update to the latest version r85 at some point, but this is good enough right now to allow testing with \`ili.


emperor/support_files/vendor/js/three.js-plugins/CanvasRenderer.js
```diff
@@ -112,11 +113,18 @@ THREE.CanvasRenderer = function ( parameters ) {
 	_normal = new THREE.Vector3(),
 	_normalViewMatrix = new THREE.Matrix3();
 
+	/* TODO
```
antgonza: Are these actually required?

ElDeveloper: I don't think they are, but if left here merging future patches to this file (from upstream) should be easier.

antgonza: do we need issues to track these down?

ElDeveloper: These are issues of three.js so I would say no, and to the best of my knowledge these issues do not restrict any of the functionality we offer.

PR #587, opened 2017-05-09
ENH: Jackknifed Plots in new-api

ElDeveloper:
This PR adds the ability to create jackknifed plots based on multiple `OrdinationResults` objects. Users can now pass a new argument to the constructor, this argument takes as an input a list of `OrdinationResults` objects, performs a series of validations, and then computes the volumes of the ellipsoids.

In order to do this I had to go through two big refactors:

* Refactoring `DecompositionModel` to be initialized with a dictionary instead of with arguments, this makes optional arguments more natural. After doing this I had to fix a large number of failing JavaScript tests, hence the big number of changes.

* Refactoring of `Emperor.make_emperor`, instead of rewriting everything I resolved that reusing `preprocess_coordinates_file` from the previous version of Emperor should be good enough for now. There are two new private methods that handle this conversion, etc. These are implicitly tested with `make_emperor`.

------------

Some things that are still up in the air:

* Should I add an example notebook?

* Should ellipsoids be tested in the view controllers? It's a bit redundant to test this in every view controller, but I can't figure out a better way to do this.


ElDeveloper:
For good measure, an animated gif:

jackknifed

tests/test_core.py
```diff
@@ -162,6 +186,32 @@ def test_custom_axes_missing_headers(self):
         with self.assertRaises(KeyError):
             emp.make_emperor(custom_axes=[':L'])
 
+    def test_jackkifed_bad_data(self):
```
antgonza: jackkifed -> jackknifed?

ElDeveloper:
@antgonza, I've addressed all your comments, I've added a few more tests, fixed failing tests, and added a new example notebook. As for reworking the underlying source, I think might be outside of the scope of this PR, and there's already plenty of changes here.

I've also taken this opportunity to cleanup the tests in Python.

ElDeveloper:
Thanks @antgonza!

ebolyen:
🎉

PR #596, opened 2017-07-21
ENH: Add opacity controller and reorder tabs

ElDeveloper:
Adds an opacity controller, and the ability to set opacity by a numeric category. Do not merge, there's still tests that need to be added and maybe some better subclassing of the OpacityViewController as it derives most of its features from ScaleViewController.

cc @qiyunzhu


ElDeveloper:
Should be ready for review now. 👍

ElDeveloper:
This appears as a big PR, but is mostly moving code from one place to the other i.e. from the ScaleController to an abstract base class (ScalarController, I know the names are terrible), same thing for the test code. This abstract class (ScalarController) is then used to create the ScaleController and the OpacityController. By this means both of the tabs act in a very similar fashion.

This is also ready for review/merge.

ElDeveloper:
Thanks @antgonza !!!

PR #600, opened 2017-08-24
Welcome to flatland (add 2D plots)

### LINKED ISSUE #303: 2D PCoA plot feature?

ElDeveloper:
Adds the ability to hide the third axis of an ordination. This can be achieved by going to the Axes tab, clicking in the dropdown menu for the third axis and selecting "Hide". When this happens the elements on screen are relocated to all lie in 0 in the Z axis. If ellipsoids are present, these are flattened. Animated gif of the UI changes:

costello-2d

As a side note, this PR also transitions the camera from a [Perspective Camera](https://en.wikipedia.org/wiki/Perspective_(graphical)) into an [Orthographic Camera](https://en.wikipedia.org/wiki/Orthographic_projection). This will help with plots included in papers, as the pictures will look flatter. Comparison of two plots showing the "keyboard dataset", note both are 3D plots:

## Perspective Camera (3D)
screen shot 2017-08-24 at 12 53 19 am

## Orthographic Camera (3D)
screen shot 2017-08-24 at 12 53 26 am

CCing @josenavas as he came up with the original idea for the implementation 👍 

Fixes #303


mortonjt:
This looks promising!  Minor comment - will the 3rd axis label be hidden in the 2D plot?  (I still Axis3 in the 2D screenshot).

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -338,20 +349,44 @@ define([
 
     // shortcut to the index of the visible dimension and the range object
     var x = this.visibleDimensions[0], y = this.visibleDimensions[1],
-        z = this.visibleDimensions[2], range = this.dimensionRanges;
+        z = this.visibleDimensions[2], range = this.dimensionRanges,
+        is2D = z === null, axesPadding = 1.07;
```
wasade: 1.07?

ElDeveloper: This adds some padding so that axes don't overlap with sample spheres.

ElDeveloper:
> Code looks good. I just have a question. I can see where the ortographic camera is being set up, however, I'm not sure I can see where the perspective camera is set up (to go from 2D to 3D). Is this because by default is set up to 3D and then later changed to 2D if needed?

@josenavas I've removed the Perspective Camera altogether, I think keeping the Orthographic camera only makes sense. But I am definitely open for suggestions and feedback on this, any thoughts?

> This looks promising! Minor comment - will the 3rd axis label be hidden in the 2D plot? (I still Axis3 in the 2D screenshot).

@mortonjt I didn't make that very clear in the text above, the screen captures show two 3D plots, the animated GIF at the beginning of the PR text is the only one that shows a 2D plot, in that case there's no axis label for the third axis.

Thanks everyone for having a look!

josenavas:
Oooh I see! Usually with the Orthographic Camera you loose the sense of depth, but in the case of emperor it was already hard to see it sometimes unless you moved the camera around. Based on your GIF, it looks like the Orthographic Camera works just fine. 🍻  👍

ElDeveloper:
> Usually with the Orthographic Camera you loose the sense of depth, but in the case of emperor it was already hard to see it sometimes unless you moved the camera around.

@josenavas Yeah, that's the main reason why I chose to go with the orthographic camera, the perspective view made it for some awkward views of the data.

> Based on your GIF, it looks like the Orthographic Camera works just fine. 🍻 👍

Cheers! 🍺

emperor/support_files/js/sceneplotview3d.js
antgonza: Would you mind adding some text/comment about that selection and if possible how it was selected? 1.07 is pretty specific.

ElDeveloper:
Should be fixed!

ElDeveloper:
Tests are passing! Woooo!

Yoshiki Vázquez-Baeza

ElDeveloper:
Thanks @antgonza!

PR #602, opened 2017-08-28
Update three.js and bug fixes

ElDeveloper:
Update all the THREE.js dependencies to the latest release. The old
version had a bug where once you changed the background color
to something that's not black, you would never be able to change it
back to black (very strange).

In addition, as I worked on the problem above, I noticed that the code was mixing color representations, as hexadecimal numbers and as strings. Which lead to the AxesViewController to stop working when colors were saved as numbers and then tried to be loaded as strings. This patch changes all properties to always deal with CSS-like strings.

Finally, in the latest version of three.js the renderer's clearColor property was deprecated and replaced by the scene's background property.

**Suggestions on how to review**: Only pay attention to the changes in, all the other files are updates from THREE.js:

```
 LICENSE.md                                                         |    2 +-
 emperor/support_files/js/axes-controller.js                        |   33 +-
 emperor/support_files/js/controller.js                             |   18 +-
 emperor/support_files/js/sceneplotview3d.js                        |   22 +-
 emperor/support_files/js/view.js                                   |    8 +-
 tests/javascript_tests/test_axes_controller.js                     |   16 +-
 tests/javascript_tests/test_decomposition_view.js                  |    4 +-
 tests/javascript_tests/test_sceneplotview3d.js                     |    8 +-
```


emperor/support_files/vendor/js/three.js-plugins/SVGRenderer.js
```diff
@@ -339,20 +372,21 @@ THREE.SVGRenderer = function () {
 		var scaleX = element.scale.x * _svgWidthHalf;
 		var scaleY = element.scale.y * _svgHeightHalf;
 
-		_svgNode = getRectNode( _rectCount ++ );
+		if ( material.isPointsMaterial ) {
+			scaleX *= material.size;
```
antgonza: why sometimes you add 2 new lines and some others not? BTW I'm not a fan of the double lines ...

ElDeveloper:
I forgot to mention that. While I was updating the other files in the `three.js-plugins` folder, I noticed that those files hadn't been updated in a while, and that they were not being included anywhere, so I removed them.

emperor/support_files/vendor/js/three.js-plugins/SVGRenderer.js
ElDeveloper: Most of the changes to SVGRenderer.js come from the THREE.js repo, I didn't write most of this code. There's one minor exception, I found a bug in the way it deals with transparent objects, and I added that in https://github.com/biocore/emperor/pull/602/commits/27c02250dfd63846c518e1ae88d92da8c89cf36c and I also submitted a PR to their repo (which already got merged):

https://github.com/mrdoob/three.js/pull/12071

antgonza:
k, what about the double new lines?

emperor/support_files/vendor/js/three.js-plugins/SVGRenderer.js
antgonza: got it, thanks.

ElDeveloper:
Thanks @antgonza!

PR #607, opened 2017-09-19
BUG: Avoid double/triple/quadruple clicks

### LINKED ISSUE #586: Visibility editor works awkwardly
After the changes in #558 the visibility controllers seem to be working awkwardly.

ElDeveloper:
~~**This PR depends on #606, so merge that one first.**~~

Fixes a problem with attribute editors requiring more than a click to
reactivate after changing value. For example users could change the
color of "Category A", after changing the color and then clicking on the
same "Category A" to change the color again, users would sometimes need
to click awkwardly until the editor reappeared on screen.

This bug affected all editors (color, visibility, scale and shape).

Fixes #586


antgonza:
Code looks good but is there a way to avoid having to add those lines to every tab js?

ElDeveloper:
Great question, not that I can think of. The objects where these are being added are as defined by SlickGrid, those don't follow a class-style definition like the objects in Emperor do so we can't subclass or do something clever about it 😞 .

ElDeveloper:
Thanks both, this should be good to merge then 🎉

PR #608, opened 2017-09-20
Adds a banner with number of visible elements

ElDeveloper:
Just like in the previous version of emperor, though this time the banner's color is aware of the scene's background color to update accordingly.

For example:

visible


ElDeveloper:
Thanks @antgonza!

Issue #580, opened 2017-04-14
phantomjs has been deprecated in favor of headless chrome


ElDeveloper:
The core maintainer of phantomjs has announced that he'll stop supporting phantomjs in favor of an upcoming feature of chrome, see link:
https://groups.google.com/forum/m/#!topic/phantomjs/9aI5d-LDuNE

This should not be a blocker yet, but it's useful to know about this.

thouger:
I use phantomjs in my work.I meet a lot of problems,include memory leak and crash etc.I wonder if these will be better in chrome.

PR #612, opened 2017-09-25
BUG: Reference to undeclared variable (urgent)

ElDeveloper:


ElDeveloper:
Unfortunately this is going to require a new release to make sure the next release of QIIME 2 is not affected.

antgonza:
How is it that test didn't catch it or changed due to this bug?

ElDeveloper:
Tests didn't catch this because controller.js is not executed as part of the test suite, since it implies making controller.js run under a headerless browser (which is not possible with phantom.js) this will likely be possible with #580. Either that, or resuscitating @emperor-helper.

My manual inspections failed because I was working with a modified version of the template that made this variable exist.

PR #629, opened 2017-11-29
ENH: Add ability to resize plot space and menu

ElDeveloper:
An example of the functionality:

resizable


antgonza:
Looks good, one question: in the previous version a double click will completely hide/show the menu, will that still be available?

ElDeveloper:
@antgonza, I just added this. If you can give it a shot, let me know what you think.

antgonza:
looks good but there are errors here ...

PR #630, opened 2017-11-29
ENH: Copy sample names to clipboard.

ElDeveloper:
Users can now click or double-click to copy a sample's name. If clicking
on a sample, they'll need to also click on the banner at the bottom
left. If double clicking on a sample, the name is automatically copied
to the clipboard.

### Example

First clicking on a sample then on the legend to get the text. And another example double-clicking a sample to get the text.

click-and-double-click

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -111,8 +111,10 @@ define([
      * @type {node}
      */
     this.$plotMenu = $("<div class='emperor-plot-menu'></div>");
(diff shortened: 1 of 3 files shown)


antgonza:
😮 there are errors here, also can you add some banner or something so the user knows this is available? Thanks!

ElDeveloper:
@antgonza, thanks for the heads up. Good call, I've added another tooltip to the controller and there's already a tooltip on the div. 👍

PR #635, opened 2017-12-18
Porting the animations tutorial

ElDeveloper:
- Re-adds the tutorial we had before.

- Points to updated files hosted on emperor.microbio.me. Note, I had to host a file in dropbox because of security issues i.e. emperor.microbio.me does not serve HTTPS.

- Include link to animations paper.


ElDeveloper:
Thanks @antgonza!

PR #636, opened 2017-12-20
ENH: Add exceptions and tests for missing samples

ElDeveloper:
Two cases are now explicitly checked for:

1) None of the samples in the ordination are part of the metadata.

2) Some samples in the coordinates matrix are not part of the metadata.
In this case the error can be overridden with the ingore_missing_samples
argument.

### DIFF
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -248,9 +248,10 @@ class Emperor(object):
         return display(HTML(str(self)))
 
     def _validate_metadata(self, ignore_missing_samples):
-        difference = set(self.ordination.samples.index) - set(self.mf.index)
+        ordination_samples = set(self.ordination.samples.index)
+        difference = ordination_samples - set(self.mf.index)
 
-        if len(difference) == len(self.ordination.samples):
+        if difference == ordination_samples:
             raise ValueError('None the sample identifiers match between the '
                              'metadata and the coordinates. Verify that you '
                              'are using metadata and coordinates corresponding'
@@ -261,7 +262,7 @@ class Emperor(object):
                            "file. Override this error by using the "
                            "`ignore_missing_samples` argument. Offending "
                            "samples: %s"
-                           % ', '.join([str(i) for i in difference]))
+                           % ', '.join(sorted([str(i) for i in difference])))
         elif difference and ignore_missing_samples:
             # pad the missing samples
             data = np.full((len(difference), self.mf.shape[1]),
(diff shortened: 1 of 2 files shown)


ElDeveloper:
Should help with https://github.com/qiime2/q2-emperor/issues/50

emperor/core.py
```diff
@@ -244,6 +247,34 @@ def _repr_html_(self):
 
         return display(HTML(str(self)))
 
+    def _validate_metadata(self, ignore_missing_samples):
+        difference = set(self.ordination.samples.index) - set(self.mf.index)
+
+        if len(difference) == len(self.ordination.samples):
```
antgonza: I don't really understand this comparison or the error, can you check them?

emperor/core.py
```diff
@@ -244,6 +247,34 @@ def _repr_html_(self):
...
+                             'metadata and the coordinates. Verify that you '
+                             'are using metadata and coordinates corresponding'
+                             ' to the same dataset.')
+
+        if difference and not ignore_missing_samples:
+            raise KeyError("There are samples not included in the mapping "
+                           "file. Override this error by using the "
+                           "`ignore_missing_samples` argument. Offending "
+                           "samples: %s"
+                           % ', '.join([str(i) for i in difference]))
```
antgonza: Should the difference be sorted?

emperor/core.py
ElDeveloper: After looking at this comparison, I've changed this to compare the sample names themselves instead of the length.

This error occurs when a user accidentally mismatches the metadata and the coordinates, if you have any suggestions to make the error message clearer that would help a lot.

emperor/core.py
ElDeveloper: Great idea!

emperor/core.py
```diff
@@ -244,6 +247,35 @@ def _repr_html_(self):
 
         return display(HTML(str(self)))
 
+    def _validate_metadata(self, ignore_missing_samples):
+        ordination_samples = set(self.ordination.samples.index)
+        difference = ordination_samples - set(self.mf.index)
+
+        if difference == ordination_samples:
+            raise ValueError('None the sample identifiers match between the '
```
antgonza: None of the ...

ElDeveloper:
Yes, the code is assuming that samples in the coordinates file are a subset of the samples  in the metadata. The other way around is not recommended but can be bypassed with the `ignore_missing_samples` argument (except when there's a total mismatch of both coordinates and metadata).

ElDeveloper:
Thanks @antgonza!

PR #637, opened 2017-12-21
BLD: Fix problem with setup.py and setuptools

### LINKED ISSUE #631: setup.py broken (by use of set type for reqs?)
Steps to reproduce:

  - Using Python 2.7.12 or 3.6.3:

```bash
virtualenv venv
source venv/bin/activate
pip install emperor
```

Output:
```
Collecting emperor
  Downloading emperor-0.9.60.tar.gz (4.1MB)
    100% |████████████████████████████████| 4.1MB 304kB/s 
    Complete output from command python setup.py egg_info:
    error in emperor setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Unordered types are not allowed
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-_Fn1vc/emperor/
```

As far as I can tell this appears to be because `install_requires` is of `set` type, which, as the error message says, is not allowed here. Changing it to a list should probably fix it? The add/remove code would presumably also need changing. (I'm guessing that the easy of add/remove is why someone used a set, but it seems weird that they presumably altered and then... didn't test it?)

ElDeveloper:
Fixes #631, thanks to @ikirker for the report and the suggested patch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,17 +42,17 @@ Gigascience. 2013 Nov 26;2(1):16.
 
 skbio_2 = "scikit-bio >= 0.4.1, < 0.5.0"
 skbio_3 = "scikit-bio >= 0.4.1"
-base = {"numpy >= 1.7", "scipy >= 0.17.0", "click", "pandas",
-        skbio_2, "jinja2 >= 2.9", "future"}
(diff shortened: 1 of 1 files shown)


ElDeveloper:
Thanks @antgonza!

PR #638, opened 2018-01-31
ENH:Download colormap instead of discrete legend

### LINKED ISSUE #336: When creating a legend using a painting that is continuous...
...it would be nice to have the gradient depicted in the legend, instead of representing the gradient as categorical.

ElDeveloper:
Previously when a continuous colormap was set and the user downloaded
the SVG image, Emperor would produce a categorical list of the values,
this PR changes this so that a colormap is produced instead.

Fixes #336


antgonza:
Code looks good but there are errors in the python 2.7 builds, seem precision errors.

ElDeveloper:
Tests should be fixed now.

PR #643, opened 2018-02-22
ENH: Warn users when hiding samples from a plot

### LINKED ISSUE #589: Warn users about changes to visibility
Users should be warned when they hide samples from a plot since this modifies the meaning of the ordination. We discussed some options in #579, for convenience these are listed here:

1) Showing a warning from the Python interface whenever the `visibility_by` method is called.
2) Show a [modal dialog](https://jqueryui.com/dialog/#modal-message) in the GUI **every single time** the visibility of a set of samples is modified.
3) Show a [modal dialog](https://jqueryui.com/dialog/#modal-message) in the GUI **only the first time** the visibility of a set of samples is modified.
4) Show a watermark on the plot anytime the visibility of a set of samples is modified. If implementing this option, we will need a way to disable the watermark for use-cases when hiding samples is not done maliciously, but we think that an appropriate requirement is that the JavaScript code be modified directly (perhaps a harsh but fair requirement).

ElDeveloper:
This patch adds a message (in the sample counter) stating that hiding
samples in an ordination can be misleading.

Fixes #589 

Example animated gif:

emperor

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -418,7 +418,7 @@ define([
    */
   EmperorController.prototype.updatePlotBanner = function() {
     var color = this.sceneViews[0].scene.background.clone(), visible = 0,
-        total = 0;
+        total = 0, message = '';
 
     // invert the color so it's visible regardless of the background
     color.setRGB((Math.floor(color.r * 255) ^ 0xFF) / 255,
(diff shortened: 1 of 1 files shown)


emperor/support_files/js/controller.js
```diff
@@ -435,7 +435,13 @@ define([
     });
 
     this.$plotBanner.css({'color': color, 'border-color': color});
-    this.$plotBanner.html(visible + '/' + total + ' visible');
+
```
antgonza: Should there be a test for this?

ElDeveloper: Unfortunately not now, as we don't have a good way to test the `EmperorController` object.

ElDeveloper:
Thanks @antgonza

Issue #634, opened 2017-12-18
saving _all_ labels


antgonza:
Currently, when saving an image + labels, only the color menu is saved but it will be great to save all options/menus that were changed in the plot; for example, shape.

ElDeveloper:
This should also include procrustes.

PR #646, opened 2018-03-09
Habemus arrow-based biplots 

ElDeveloper:
# TL;DR

Arrow-based biplots work 🎉 
We need feedback 🙏 
This is a big changeset 🤕 

cc @antgonza @mortonjt @wasade 

----------------------

This pull request adds the ability for the `Emperor` class to display biplots based on the `OrdinationResults` class. This resulted in one visible change to the Python API, the `Emperor` class can now receive as a parameter feature metadata. The metadata is validated just like with sample metadata. Since feature metadata has traditionally not been as copious and readily available, `Emperor` also works when no feature metadata is present.

Internally, some methods were changed to be made feature- or sample-agnostic. And beyond this, I simplified the serialization of objects so that only a dictionary containing all the data is ever serialized, see the `_to_dict` method.

In the JavaScript API, the `DecompositionModel` class gained a new attribute, `type`. This attribute indicates whether the decomposition is a scatter type (normal plots) or an arrow type (biplots). Depending on the type the `DecompositionView` and the view controllers perform different actions. This makes it such that all controllers (that make sense) can operate on both decompositions seamlessly.

Further down the internals of the display logic, we use arrows to show biplots, these are created with a custom subclass of THREE's `ArrowHelper` class (`EmperorArrowHelper`). These objects can be interacted with just like with spheres, etc.

Through this changeset you will notice a number of changes in "random" spots, these were to account for modifications needed to truly operate on multiple decompositions.

All the menus in the attribute tabs will show metadata names under the 'biplot' name (scroll to the bottom on the dropdown). Anything you can do with a sample, you can do with an arrow, except change shape or scale as it would change the meaning of the plot itself.

### Examples

I've included a notebook that should create a live biplot so you can try this out. **Please** try this out as the functionality is rather new.

**Color and manipulate the arrows based on a feature metadata category**:

biplot-3

If you are curious about what happens with no metadata, unfold the gif below:

<details>
biplot-4
</details>

### Feedback and open items

- [x] @antgonza mentioned that it would be nice to show labels on the arrows indicating the name of the feature. We tried to think about a number of alternatives, but couldn't quite come up with something good other than labels. While the metadata can be displayed and the arrows colored, this is probably not sufficient in general. Also, labels were historically not great because they would often be rather long strings, and it was a bit hard to make anything out of them when there were more than a handful, trimming them helped but it wasn't quite the most useful thing.

- [x] ~Saving labels is a bit fuzzy right now, this would be better addressed in #634. This would be done by implementing a method for each controller `makeLegend` that way the controller can take care of any other necessary business~.

- [x] ~Saving settings is also rather fuzzy~.

- [x] ~The controllers need to implement a cache to be able to switch back and forth between categories without needing to reset custom set values~.


emperor/core.py
```diff
@@ -199,15 +205,24 @@ class Emperor(object):
...
 
-        self._validate_metadata(ignore_missing_samples)
+        self.mf = self._validate_metadata(self.mf, self.ordination.samples,
+                                          ignore_missing_samples)
+
+        # if biplots are to be visualized
+        if self.ordination.features is not None:
+            self.feature_mf = self._validate_metadata(feature_mapping_file,
+                                                      self.ordination.features,
+                                                      False)
```
wasade: would it be possible to associate the kwarg with the `False` for context?

emperor/core.py
```diff
@@ -377,29 +401,77 @@ def make_emperor(self, standalone=False, custom_axes=None,
...
+            The output of _process_data. Should contain information about the
+            scatter plot and the biplot.
+
+        Returns
+        -------
+        dict
+            A dictionary describing the plots contained in the ordination
+            object and the sample + feature metadata.
+        """
+        # data is a tuple as returned by _process_data
```
wasade: Would it make sense for `_process_data` to return a `namedtuple`?

emperor/core.py
```diff
@@ -514,14 +587,26 @@ def _process_data(self, custom_axes, jackknifing_method):
...
+                bi_headers, bi_metadata)
 
-    def _to_legacy_map(self, custom_axes=None):
+    def _to_legacy_map(self, mf, custom_axes=None):
         """Helper method to convert Pandas dataframe to legacy QIIME structure
 
         Parameters
         ----------
+        mf : pd.DataFrame
+            DataFrame with the metadata, this can be feature or sample
```
wasade: Does the index matter for this dataframe?

emperor/support_files/js/draw.js
```diff
@@ -77,6 +112,37 @@ define(['underscore', 'three'], function(_, THREE) {
...
+   * originates from.
+   * @param {float[]} to The x, y and z coordinates where the arrow points to.
+   * @param {integer} color Hexadecimal base that specifies the color of the
+   * line.
+   *
+   * @return {THREE.Object3D}
+   * @function makeArrow
+   */
+  function makeArrow(from, to, color) {
+    // TODO: Write tests for this function
```
wasade: Is this something you'd like to do for this PR, or should this be noted as an issue?

emperor/support_files/js/opacity-view-controller.js
```diff
@@ -52,10 +52,21 @@ define([
...
-      scope.markers[element.idx].material.opacity = opacity;
-    });
+    if (scope.decomp.isScatterType()) {
+      _.each(group, function(element) {
+        scope.markers[element.idx].material.transparent = transparent;
+        scope.markers[element.idx].material.opacity = opacity;
+      });
+    }
+    else if (scope.decomp.isArrowType()) {
+      _.each(group, function(element) {
```
wasade: Would it make sense to pull these anonymous functions out? It has the benefit of simplifying the if/elseif by reducing the LOC of each branch

emperor/support_files/js/opacity-view-controller.js
```diff
@@ -72,10 +83,21 @@ define([
 
     // webgl acts up with transparent objects, so we only set them to be
     // explicitly transparent if the opacity is not at full
-    _.each(dv.markers, function(element) {
-      element.material.transparent = transparent;
-      element.material.opacity = value;
-    });
+    if (dv.decomp.isScatterType()) {
+      _.each(dv.markers, function(element) {
```
wasade: similar comment here w.r.t. the anonymous function

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -685,10 +685,13 @@ define([
       this.control.update();
     }
 
-    //point all samples towards the camera
-    _.each(this.decViews.scatter.markers, function(element) {
-      element.quaternion.copy(camera.quaternion);
-    });
+    // don't point arrows towards the camera, that makes zero sense :L!
+    if (this.decViews.scatter.decomp.isScatterType()) {
+      //point all samples towards the camera
```
wasade: this comment seems to contradict the prior...

emperor/support_files/js/view.js
```diff
@@ -101,37 +103,55 @@ DecompositionView.prototype._initBaseView = function() {
 
   hasConfidenceIntervals = this.decomp.hasConfidenceIntervals();
 
-  this.decomp.apply(function(plottable) {
-    mesh = new THREE.Mesh(geometry, new THREE.MeshPhongMaterial());
-    mesh.name = plottable.name;
+  if (this.decomp.isScatterType()) {
```
wasade: there is quite a bit of work going on within this if-statement, would it make sense to decompose?

emperor/support_files/templates/logic-template.html
```diff
@@ -101,35 +101,26 @@
...
 
-  var dm, ec;
+  var plot, biplot = null, ec;
 
   function init() {
-    var decomp = data['decompositions'][0];
+    // Initialize the DecompositionModel for the scatter plot, and optionally
+    // add one for the biplot arrows
+    plot = new DecompositionModel(data.plot.decomposition,
+                                     data.plot.metadata_headers,
```
wasade: the spacing here is a little off

tests/test_core.py
```diff
@@ -56,6 +56,27 @@ def setUp(self):
...
+                              proportion_explained=props.copy(),
+                              short_method_name='RDA',
+                              long_method_name='Redundancy Analysis')
+
+        self.biplot = d
+
+        feature_mf = pd.DataFrame(index=['f.PC.636', 'f.PC.635', 'f.PC.356',
+                                         'f.PC.481', 'f.PC.354'])
+        feature_mf['Category'] = ['foo', 'bar', 'baz', 'foo', 'foo']
+        feature_mf['Second'] = ['No', 'Yes', 'Noes', 'Noooo', 'Yep']
```
wasade: Can the feature data be continuous?

emperor/core.py
ElDeveloper: For sure, added.

emperor/core.py
ElDeveloper: I don't think so, `_process_data` is a private method not intended for use by anything else but `_to_json`.

emperor/core.py
ElDeveloper: It doesn't really matter, I've expanded on the description.

emperor/support_files/js/draw.js
ElDeveloper: I've removed this comment, there are now tests for this function 👍

emperor/support_files/js/opacity-view-controller.js
ElDeveloper: This is a great suggestion. I think the code is much more readable now 👍

emperor/support_files/js/opacity-view-controller.js
ElDeveloper: Fixed with the approach you described above.

emperor/support_files/js/sceneplotview3d.js
ElDeveloper: Good catch, thanks. I've rephrased these comments for them to make sense.

emperor/support_files/js/view.js
ElDeveloper: Yes, I also noticed this and even worse there's some duplicated code with this and other method. I've refactored this in #647

emperor/support_files/templates/logic-template.html
ElDeveloper: Fixes

tests/test_core.py
ElDeveloper: Yes it can :)

mortonjt:
One thing that would be extremely useful would be to have some helper code / tutorials that show how to feed in inputs to the biplots.

The biplots at the end of the day need have an Ordination results object.

The way that I typically generate biplots is to perform SVD and feed in the `res.samples` and `res.features` as follows (using compositional PCA)

```python
import numpy as np
from skbio.stats.composition import clr, centralize
mat = clr(centralize(closure(table + 1)))
u,k,v = np.linalg.svd(mat)
n = len(u)
_k = k[:2]
u = u[:,:2] * np.sqrt(n-1)  # scale U matrix wrt to sqrt of eigenvalues
v = np.multiply(v[:2,:],(_k.reshape(2,1) / np.sqrt(n-1)))   # scale V matrix wrt to sqrt of eigenvalues

u = pd.DataFrame(u, index=<sample-ids>, columns=<PC axes>)
v = pd.DataFrame(v.T, index=<feature-ids>, columns=<PC axes>)
res = OrdinationResults(...)
res.samples = u
res.feature = v
# supply metadata `mf` and `feature_mf`
Emperor(res, mf, feature_mf, remote=False)
```
A similar procedure can be done using multomics with PLSSVD (see office notebook [here](https://github.com/knightlab-analyses/office-study/blob/master/ipynb/plssvd.ipynb)).  We will eventually want to have a tutorial that shows how to do this.

Doing this with PCoA isn't as straightforward -- the only way that I'm aware that does this is using weight-averages from Dan's code in qiime1 [here](https://github.com/biocore/qiime/blob/master/qiime/biplots.py).  I don't entirely understand the math behind it, but it may be worthwhile to have a tutorial using this approach as well.

ElDeveloper:
Thanks for reviewing @mortonjt and @wasade!

> One thing that would be extremely useful would be to have some helper code / tutorials that show how to feed in inputs to the biplots.
>
> The biplots at the end of the day need have an Ordination results object.
>
> The way that I typically generate biplots is to perform SVD and feed in the res.samples and res.features as follows (using compositional PCA).

Thanks Jamie, that's a great idea, I'll add that code to the biplots.ipynb notebook. It will be better to showcase this code instead of the random feature generation that's currently in place.

One question here, does anyone have a dataset that they think might be specifically fitting here?

> We will eventually want to have a tutorial that shows how to do this.

I agree. While chatting with Jamie we think this makes more sense to put in QIIME 2 as a plugin that can then be used together with q2-emperor.

👍

mortonjt:
I think the biplots in the PTSD example is a really good one to try out
first, particularly at the genus level.   The dataset can be found here:
https://qiita.ucsd.edu/study/description/1634

Performing PLSSVD on multiomics is a bit tricker, since there is needs a
bit of data massaging to classify both sets of features.
The office study datasets may be a good choice, especially since the class
assignments are really obvious.  But I'd try reproducing the PTSD biplots
first, since it is a smaller / easier dataset.
Office study datasets can be found below
https://qiita.ucsd.edu/study/description/11079
https://qiita.ucsd.edu/study/description/10244

ElDeveloper:
Thanks so much for these pointers, this is very helpful!

mortonjt:
Sickkkkk

The only thing that I would comment on is that it would be great if in the tutorial, the `Day` column was highlighted as a gradient, so that the differences between samples could be readily visualized.

Otherwise, I think this is basically good to go.  This is going to be awesome!!!

ElDeveloper:
@mortonjt, I just added the comment you requested. Here's how it looks:

biplot-colors

mortonjt:
Thanks @ElDeveloper !  This is awesome!

ElDeveloper:
Thanks the reviews and help @mortonjt, @wasade and @antgonza.

PR #647, opened 2018-03-16
Procrustes new api

ElDeveloper:
### ~~Depends on #646 (merge and review that one first).~~

This PR adds procrustes plots to the Emperor object, the interface works pretty much the same as the jackknifed plots.

These changes should also make some additional performance improvements compared to the previous version. Namely, the lines themselves are grouped into a single object. This helps the renderer draw these elements faster (see the BufferGeometry), and the lines all share a material. This is important because some small Procrustes plots can have a few thousand lines.

In addition, this also adds a column (`__Procrustes_Names__`) to the mapping file that describes each ordination (according to some annotations). This is helpful when you compare more than two matrices. For example in the notebook in this PR I compare 3 different metrics, see the animated gif:

procrustes


emperor/support_files/js/axes-controller.js
```diff
@@ -428,6 +458,146 @@ define([
...
+    var out = '';
+    var shouldDraw = _.values(this.decompViewDict).some(function(view) {
+      return view.decomp.edges.length > 0;
+    });
+
+    // if we have at least one decomposition with edges then we add the
+    // controllers.
+    if (shouldDraw) {
+      out += '<tr><td>&nbsp;</td></tr>';
+      out += '<tr><td>Edge Color (reference)</td>';
```
antgonza: missing `</td>`

emperor/support_files/js/axes-controller.js
```diff
@@ -428,6 +458,146 @@ define([
...
+      out += '<td><input type="text" name="reference-edge-color"/></td></tr>';
+      out += '<tr><td>Edge Color (other)</td>';
```
antgonza: missing `</tr>`

emperor/support_files/js/axes-controller.js
```diff
@@ -428,6 +458,146 @@ define([
...
+      out += '<td><input type="text" name="other-edge-color"/></td>';
```
antgonza: does it need `<tr></tr>`

emperor/support_files/js/controller.js
```diff
@@ -505,7 +505,7 @@ define([
 
     // this is the number of expected view controllers that will announce that
     // all view controllers have been successfully loaded.
-    this._expected = 5;
+    this._expected = 7;
```
antgonza: Should this be self calculated? Seems hard to have to chance as controllers are added ...

ElDeveloper:
> I think at least another reviewer and perhaps another 2 will be good for this PR, due to size.

I agree, **if there's another kind soul listening** I would appreciate a review. My recommendation is to pull down this branch and try out the functionality as included in the procrustes.ipynb notebook. That will provide you with a better idea of what's going on in the code.

📣 👂 :shipit:

mortonjt:
Looks good to me

ElDeveloper:
Unless anyone has additional comments, this PR should be ready to be merged.

PR #653, opened 2018-03-28
BUG: Refresh button state

ElDeveloper:
According to the documentation[1], the UI has to be manually refreshed
after changes to a button's state.

[1] - http://api.jqueryui.com/button/ see the refresh() section


ElDeveloper:
Thanks @antgonza!

PR #662, opened 2018-04-12
BUG: Hide edges when hiding visible samples

### LINKED ISSUE #193: able to hide both dots and edges of a given group in a procrustes plot
currently only dots can be hidden, leaving their edges on the plot. It would be great to provide the ability to hide both of them to keep the rest of figure visually clean.

ElDeveloper:
For procrustes plots edges were not being hidden when samples in it were. Demo:

procrustes-bug

Fixes #193

### DIFF
--- a/emperor/support_files/js/view.js
+++ b/emperor/support_files/js/view.js
@@ -258,15 +258,35 @@ DecompositionView.prototype.updatePositions = function() {
   // edges are made using THREE.LineSegments and a buffer geometry so updating
   // the position takes a bit more work but these objects will render faster
   if (this.decomp.edges.length) {
-    var left, center, right, u, v, positionsLeft, positionsRight, j = 0;
-    positionsLeft = this.lines.left.geometry.attributes.position.array;
-    positionsRight = this.lines.right.geometry.attributes.position.array;
+    this._redrawEdges();
+  }
+  this.needsUpdate = true;
+};
 
-    this.decomp.edges.forEach(function(edge) {
-      u = edge[0];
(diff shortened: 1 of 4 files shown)


emperor/support_files/js/view.js
```diff
@@ -414,5 +435,70 @@ DecompositionView.prototype.setCategory = function(attributes,
...
+
+  hideAll = plottables === undefined;
+
+  this.decomp.edges.forEach(function(edge) {
+    u = edge[0];
+    v = edge[1];
+
+    if (hideAll ||
+        (plottables.indexOf(u) !== -1 || plottables.indexOf(v) !== -1)) {
+      positionsLeft[(j * 6)] = 0;
```
antgonza: The 1->5 sounds/looks like a variable or something like that ... or why 5?

antgonza: Thus, this is a global parameter defined by BufferGeometry and if changed this will change, right? Therefore, should we have it encapsulated somehow (method?) so we don't need to change multiple places if we move to something different? If the answer is yes and the changes are complex, opening an issue is fine.

ElDeveloper: Fair enough, I've created a wrapper method for these changes. 👍

antgonza:
Great, thanks. Just to confirm, that kind of modification is not happening anywhere else in the code? What I mean, this new code can't be use somewhere else?

ElDeveloper:
Just double checked and those are the only places where the
modifications are happening.

PR #673, opened 2018-05-07
Use shaders to render hundreds of thousands of samples with a responsive UI

ElDeveloper:
## TL;DR: here's a responsive plot with [all the public samples from Qiita](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdl.dropbox.com%2Fs%2Fp6j6ydlicq9tc53%2Fgotta-go-fast.qzv%3Fdl%3D1).

This PR adds a **fast way** 🔥  to render large datasets and make the UI responsive. In order to do this we move away from using individual geometries per sample and instead use a point cloud or "particle cloud" to render all the samples in a single geometry. As it stands this mode will automatically activate when there's more than 20,000 samples (we can change this to something else).

To do this I implemented a custom shader that draws each sample as a circle, overall this works well since the density of data with large datasets doesn't require a high-level of detail per sample.

By using shaders we are limited on a few respects:
- The shapes controller is disabled, since drawing custom shapes would require a fair amount of shader code.
- Exporting as ~~PNG or~~ SVG is currently not supported since we are using a custom shader and shaders are not supported by ~~these renderers~~ this renderer.
- We can't use jackknifed plots.

While this PR is large, most of the changes are test code. Most of the relevant changes are in view.js. I implemented a few convenience setters for the visual attributes (color, size, etc) and added `_fastInit` to create the shader-based geometry.

---------

The ordination is courtesy of @wasade 🎩


antgonza:
😮 with the example!! 🎆

emperor/support_files/js/view.js
```diff
@@ -79,8 +81,22 @@ function DecompositionView(decomp) {
    */
   this.lines = {'left': null, 'right': null};
 
+  /**
+   * Whether or not the view relies on a PointCloud to display the data.
+   * @type {Bool}
+   */
+  Object.defineProperty(this, 'usesPointCloud', {
+    value: (this.decomp.length > 20000) || asPointCloud,
```
wasade: it might be nice to be able to override this threshold?

emperor/support_files/js/view.js
```diff
@@ -89,6 +105,18 @@ function DecompositionView(decomp) {
   this.needsUpdate = true;
 }
 
+/**
+ * Calculate the appropriate size for a geometry based on the first dimension's
+ * range.
+ */
+DecompositionView.prototype.getGeometryFactor = function() {
+  // this is a heauristic tested on numerous plots since 2013, based off of
```
wasade: heuristic

emperor/support_files/js/view.js
ElDeveloper: It can be overridden with the `asPointCloud` boolean value. This value can be changed from the constructor of this class. I originally added this to make testing quicker, but in general can be useful for other purposes.

antgonza: Is a plot with 20K slow and terrible? Just wondering how was this number selected?

antgonza:
BTW the save fig png/svg doesn't work

ElDeveloper:
> Is a plot with 20K slow and terrible? Just wondering how was this number selected?

I chose this arbitrarily, roughly based on the size of the American Gut and EMP (which originally motivated this feature). To be fair, I don't really know what's an appropriate size or how to determine that this *fast mode* should be used.

Do you have any suggestions?

> BTW the save fig png/svg doesn't work

Yes, this is one of the limitations noted above. Using shaders won't allow for any of the Canvas or SVG renderers to work, since the geometry is custom made. Perhaps the controls to do this should be disabled, just like how the shapes controller is not even instantiated. What do you think?

ElDeveloper:
> 20K: I guess we can go with this and change as we learn more about it

Sounds good to me, thanks!

> Export: disable and add a message so users don't get confused

Just added this, looks like this (note this is a small dataset I forced to use "fast mode"):

disabling-demo

ElDeveloper:
Great point @josenavas, while I couldn't find a way to get access to the original GL buffer, I switched to take the rasterized scene that's already present in the DOM (it has a few limitations) but it gets the job done 👍 .

ElDeveloper:
Thanks! 🎆

PR #676, opened 2018-05-31
ENH: Add menu item to hide biplot labels

ElDeveloper:
The labels can get really crowded if a lot of features are shown on screen. This PR adds a menu item to hide arrow labels, demo:

labels

**Note**: I've noticed that there's a problem with the label size which is obvious for taxonomic names (see in the image above), I'm working on a fix but don't really anything yet.

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -580,6 +580,13 @@ define([
             });
           }
         },
+        'labels' : {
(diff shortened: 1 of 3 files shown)


ElDeveloper:
This currently needs to be saved/loaded in the settings. So don't merge until that is added.

ElDeveloper:
Granted that tests pass, this should be ready for review/merge.

PR #693, opened 2018-09-12
BLD: General updates to the repository

ElDeveloper:
- Updates the setup.py file to no longer use distutils.
- Improves the long_description section to load the README.md file.
- Adds publications to the README.md file


ElDeveloper:
For reference, here's how the [current page looks like in PYPI](https://pypi.org/project/emperor/) and [here's a link to the new page](https://test.pypi.org/project/emperor/) in test.pypi.org.

.travis.yml
```diff
@@ -1,36 +1,57 @@
...
+  # https://github.com/travis-ci/travis-ci/issues/9815
+  include:
+    - python: 3.7
+      dist: xenial
+      sudo: true
+      env: NUMPY_VERSION=""
+    - python: 3.7
+      dist: xenial
+      sudo: true
+      env: NUMPY_VERSION="<1.13"
```
mortonjt: Is there any reason why this is only tested against numpy<1.13?  That version of numpy is about 2 years out of date ...

mortonjt:
Makes sense to me -- but it would be great to see numpy get updated first

.travis.yml
ElDeveloper: For now it is mostly because we have compatibility and we don't really use any new features. Also, Emperor is compatible with this and with much newer versions, hence the other builds with `NUMPY_VERSION=""`.

ElDeveloper:
> Makes sense to me -- but it would be great to see numpy get updated first

Numpy is up to date though 👍

ElDeveloper:
Can this be reviewed/merged?

ElDeveloper:
Thanks!

PR #694, opened 2018-09-13
BUG: Fix version conflicts with Jupyter deps

ElDeveloper:
Fixes a problem where Jupyter's require declarations would shadow
Emperor's. Specifically while using the newest version of the Jupyter
notebook jquery would be updated to version 3.x and that version is not
compatible with several dependencies of Emperor.

For now this fix ensures that there will be no version conflicts by
putting all our dependencies under their own context.

----------

In case someone runs into this problem, here's how the UI would look (notice the dropdown menus are not loaded properly and no categories are visible):

screen shot 2018-09-13 at 4 15 08 pm

### DIFF
--- a/emperor/support_files/templates/logic-template.html
+++ b/emperor/support_files/templates/logic-template.html
@@ -4,7 +4,11 @@
 </div>
 
 <script type="text/javascript">
-requirejs.config({
+// When running in the Jupyter notebook we've encountered version conflicts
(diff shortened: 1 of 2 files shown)


antgonza:
Code looks good but there are errors ...

ElDeveloper:
Tests are passing now.

PR #702, opened 2018-11-29
DEV: Bump version strings

ElDeveloper:


tests/all_tests.py
```diff
@@ -95,7 +95,7 @@ def test(suppress_unit_tests, suppress_javascript_unit_tests, unittest_glob):
             command = '%s %s -v' % (python_name, unittest_name)
             stdout, stderr, return_value = console(command)
             print(stderr.decode("utf-8"))
-            if not unittest_good_pattern.search(stderr):
```
wasade: should the `print` above this line be removed?

ElDeveloper: The print is used to show the output of the unit tests, so it is needed.

PR #703, opened 2018-11-29
 BUG: Fix a problem with no metadata columns

### LINKED ISSUE #700: Plot fails with zero columns and just an index in the dataframe
See this forum thread:
https://forum.qiime2.org/t/emperor-plot-stuck-with-550-samples/

ElDeveloper:
Fixes a problem where passing a dataframe with no metadata columns would
lead to a failure in the UI. The problem originated from the way we were
using the apply function to transform the values to strings, and how we
were merging the index's names.

Fixes #700


ElDeveloper:
Can this be reviewed/merged?

ElDeveloper:
Thanks!

PR #706, opened 2019-01-11
Fixes problem with the camera not being centered

### LINKED ISSUE #696: Plot zoomed in and panned over when initially loaded
I was testing out @antgonza's [QIIME 2 FMT cdiff tutorial PR](https://github.com/qiime2/docs/pull/353/) and noticed a plot that looks like this when it loads:

screen shot 2018-09-28 at 10 42 53 am

This is what it looks like zoomed out:

screen shot 2018-09-28 at 10 43 32 am

I checked this on FF and Chrome. Here is the [offending QZV](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum-qiime2-org.s3.dualstack.us-west-2.amazonaws.com%2Foriginal%2F2X%2Ff%2Ff51a61dbbdcf8d1f2b68dd6fd3347c87f6d584cd.qzv), in case you want to test. Otherwise, steps to reproduce are in the PR referenced above.

ElDeveloper:
Fixes #696 

The problem was that the camera assumed all coordinate spaces were centered around the origin. This assumption is not always true and it is **almost always** not true for weighted UniFrac (hence the panning of the camera in the screenshot in #696). While fixing this, I noticed that the target of the camera had an offset making plots rotate in an awkward way, this was fixed by adding calculating the target.

I also noticed that "re-centering" wasn't working properly either, re-centering the camera would lead to "almost" centered scenes. This last problem was fixed by disabling inertia/damping in the scene's controls.

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -332,7 +332,7 @@ define([
     }
 
     var spv = new ScenePlotView3D(this.renderer, this.decViews,
-                                  this.$plotSpace, 0, 0, 0, 0);
+                                  this.$plotSpace, 0, 0, this.width, this.height);
     this.sceneViews.push(spv);
 
     // this will setup the appropriate sizes and widths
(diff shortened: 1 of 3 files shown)


thermokarst:
❤️ - thanks @ElDeveloper!

PR #707, opened 2019-01-15
BUG: Fix problem with two-dimensional data

### LINKED ISSUE #686: Emperor silently fails if only two dimensions are provided
See [this forum post](https://forum.qiime2.org/t/problem-viewing-emperor-plot-with-qiime-tools-view-command-and-view-qiime2-org).

### LINKED ISSUE #705: Silent error with 2 samples
Looks like if 2 samples are being compared, there's a silent error happening in the UI. See this forum post:
https://forum.qiime2.org/t/emperor-qiime2-2018/7417/3

ElDeveloper:
If the plot was initialized with two dimensions, the UI would fail to
instantiate as several objects had the dimensions hard-coded. I've
changed this to make it dependant on the data. In addition the Python
API will now check for one-dimensional data.

Fixes #686
Fixes #705

### DIFF
--- a/emperor/support_files/js/view.js
+++ b/emperor/support_files/js/view.js
@@ -37,16 +37,19 @@ function DecompositionView(decomp, asPointCloud) {
   /**
    * Top visible dimensions
    * @type {integer[]}
-   * @default [0, 1, 2]
    */
-  this.visibleDimensions = [0, 1, 2]; // We default to the first three PCs
+  // this array should have at most three elements
+  this.visibleDimensions = _.range(this.decomp.dimensions).slice(0, 3);
   /**
    * Orientation of the axes, `-1` means the axis is flipped, `1` means the
    * axis is not flipped.
    * @type {integer[]}
-   * @default [1, 1, 1]
    */
-  this.axesOrientation = [1, 1, 1];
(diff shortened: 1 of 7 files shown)


emperor/support_files/js/view.js
```diff
@@ -37,16 +37,19 @@ function DecompositionView(decomp, asPointCloud) {
   /**
    * Top visible dimensions
    * @type {integer[]}
-   * @default [0, 1, 2]
    */
-  this.visibleDimensions = [0, 1, 2]; // We default to the first three PCs
+  // this array should have at most three elements
+  this.visibleDimensions = _.range(this.decomp.dimensions).slice(0, 3);
```
antgonza: Based on the comment, I feel like we need a if/else or perhaps I'm not clear?

ElDeveloper: The way `slice` works allows us to not need an if/else. For example if the number of dimensions is 10 `_.range` is going to produce `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]` and `slice` will reduce this to `[0, 1, 2]`. In the case where the number of dimensions is 2 (what originated this pull request) the output of `_.range` will be `[0, 1]` and `slice` will keep this is as `[0, 1]` giving the correct number of visible dimensions.

PR #724, opened 2019-07-25
Fast Animation

dhakim87:
I split the Tube rendering into static and dynamic portions.  Static tubes are computed once upon starting playback.  Dynamic tubes are recalculated each frame.  Static tube draw range is recalculated each frame as well.  The effect is that we draw all the animation up to a certain point, then fit a single interpolated segment onto the end of it.  This is much faster than recomputing all the tubes every frame.  

Smooth on 20000 points on my laptop.  Should check it on more hardware since it touches the webGL layer.

### DIFF
--- a/emperor/support_files/js/animate.js
+++ b/emperor/support_files/js/animate.js
@@ -137,7 +137,7 @@ function(_, trajectory) {
      * @default -1
      */
     this.currentFrame = -1;
-    
+
     /**
      * @type {Integer}
      * The previous frame served by the director
@@ -145,7 +145,7 @@ function(_, trajectory) {
      * but built more generically to allow rewind and frame skipping in the future)
      */
     this.previousFrame = -1;
-    
+
     /**
      * @type {Array}
      * Array where each element in the trajectory is a trajectory with the
--- a/emperor/support_files/js/animate.js
+++ b/emperor/support_files/js/animate.js
@@ -138,7 +138,7 @@ function(_, trajectory) {
      */
     this.currentFrame = -1;
     
-    /*
+    /**
      * @type {Integer}
      * The previous frame served by the director
      * (Note, more or less always equals currentFrame - 1 in current implementation,
--- a/emperor/support_files/js/animate.js
+++ b/emperor/support_files/js/animate.js
@@ -141,8 +141,6 @@ function(_, trajectory) {
     /**
      * @type {Integer}
      * The previous frame served by the director
-     * (Note, more or less always equals currentFrame - 1 in current implementation,
-     * but built more generically to allow rewind and frame skipping in the future)
      */
     this.previousFrame = -1;
 
--- a/emperor/support_files/js/animate.js
+++ b/emperor/support_files/js/animate.js
@@ -137,6 +137,15 @@ function(_, trajectory) {
      * @default -1
      */
     this.currentFrame = -1;
+    
+    /*
+     * @type {Integer}
+     * The previous frame served by the director
+     * (Note, more or less always equals currentFrame - 1 in current implementation,
+     * but built more generically to allow rewind and frame skipping in the future)
+     */
+    this.previousFrame = -1;
+    
     /**
      * @type {Array}
      * Array where each element in the trajectory is a trajectory with the
@@ -276,6 +285,7 @@ function(_, trajectory) {
    */
   AnimationDirector.prototype.updateFrame = function() {
     if (this.animationCycleFinished() === false) {
+      this.previousFrame = this.currentFrame;
       this.currentFrame = this.currentFrame + 1;
     }
   };
--- a/emperor/support_files/js/animations-controller.js
+++ b/emperor/support_files/js/animations-controller.js
@@ -459,7 +459,7 @@ define([
 
     view.staticTubes = [];
     view.dynamicTubes = [];
-   
+
     view.needsUpdate = true;
 
     this._updateButtons();
@@ -575,7 +575,7 @@ define([
         disposeTrajectoryLineDynamic(tube);
       }
     });
-    
+
     //Construct new dynamic tubes containing necessary interpolated segment for the current frame
     view.dynamicTubes = this.director.trajectories.map(function(trajectory) {
       var color = scope._colors[trajectory.metadataCategoryName] || 'red';
--- a/emperor/support_files/js/animations-controller.js
+++ b/emperor/support_files/js/animations-controller.js
@@ -443,8 +443,7 @@ define([
     this.playing = false;
     this.director = null;
 
-    var allTubes = view.staticTubes.concat(view.dynamicTubes);
-    allTubes.forEach(function(tube) {
+    view.getTubes().forEach(function(tube) {
       if (tube !== null && tube.parent !== null) {
         tube.parent.remove(tube);
       }
@@ -544,8 +543,7 @@ define([
     var radius = view.getGeometryFactor();
     radius *= 0.45 * this.getRadius();
 
-    var i = 0;
-    for (i = 0; i < this.director.trajectories.length; i++) {
+    for (var i = 0; i < this.director.trajectories.length; i++) {
       var trajectory = this.director.trajectories[i];
 
       //Ensure static tubes are constructed
--- a/emperor/support_files/js/animations-controller.js
+++ b/emperor/support_files/js/animations-controller.js
@@ -10,7 +10,6 @@ define([
 ], function($, _, DecompositionView, ViewControllers, AnimationDirector,
             draw, Color, ColorViewController) {
   var EmperorViewController = ViewControllers.EmperorViewController;
-  var drawTrajectoryLine = draw.drawTrajectoryLine;
   var drawTrajectoryLineStatic = draw.drawTrajectoryLineStatic;
   var drawTrajectoryLineDynamic = draw.drawTrajectoryLineDynamic;
   var updateStaticTrajectoryDrawRange = draw.updateStaticTrajectoryDrawRange;
--- a/emperor/support_files/js/animations-controller.js
+++ b/emperor/support_files/js/animations-controller.js
@@ -11,6 +11,9 @@ define([
             draw, Color, ColorViewController) {
   var EmperorViewController = ViewControllers.EmperorViewController;
   var drawTrajectoryLine = draw.drawTrajectoryLine;
+  var drawTrajectoryLineStatic = draw.drawTrajectoryLineStatic;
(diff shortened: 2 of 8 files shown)


emperor/support_files/js/animations-controller.js
```diff
@@ -440,13 +443,16 @@ define([
     this.playing = false;
     this.director = null;
 
-    view.tubes.forEach(function(tube) {
-      if (tube.parent !== null) {
+    var allTubes = view.staticTubes.concat(view.dynamicTubes);
```
ElDeveloper: This type of operation or pattern on both types of tubes happens a couple of times. Do you think we should have a `DecompositionView.getTubes` method that returns both types of tubes?

emperor/support_files/js/draw.js
```diff
@@ -295,11 +299,80 @@ define(['underscore', 'three', 'jquery'], function(_, THREE, $) {
     // the line will contain the two vertices and the described material
     // we increase the number of points to have a smoother transition on
     // edges i. e. where the trajectory changes the direction it is going
-    lineGeometry = new THREE.TubeGeometry(path, (points.length - 1) * 3, radius,
```
ElDeveloper: ^ I think we've stopped using the `drawTrajectoryLine` function in the code-base if that's the case can you delete that function and all other associated tests?

emperor/support_files/js/draw.js
```diff
@@ -295,11 +299,80 @@ define(['underscore', 'three', 'jquery'], function(_, THREE, $) {
...
+    return new THREE.Mesh(tubeBufferGeom, material);
+  }
+  
+  function updateStaticTrajectoryDrawRange(trajectory, currentFrame, threeMesh)
+  {
+    //Blah, reverse engineering the number of points in a THREE tube is not fun, and may be implementation/version dependent.
+    //Number of points drawn per tube segment = 2 (triangles) * 3 (points per triangle) * NUM_TUBE_CROSS_SECTION_POINTS (number of vertices in a cross section of tube)
+    //Number of tube segments per pair of consecutive points = NUM_TUBE_SEGMENTS
+
+    var multiplier = 2 * 3 * NUM_TUBE_CROSS_SECTION_POINTS * NUM_TUBE_SEGMENTS
```
ElDeveloper: I think the "parameters" property might have this information stored somewhere. Let's have a look at it on Monday.

ElDeveloper:
BTW, @antgonza if you can have a look that would be very helpful!

rob-knight:
This sounds awesome!

On Jul 26, 2019, at 6:48 PM, Yoshiki Vázquez Baeza <notifications@github.com<mailto:notifications@github.com>> wrote:


BTW, @antgonza<https://github.com/antgonza> if you can have a look that would be very helpful!

emperor/support_files/js/draw.js
```diff
@@ -295,11 +299,80 @@ define(['underscore', 'three', 'jquery'], function(_, THREE, $) {
...
+    // based on the example described in:
+    // https://github.com/mrdoob/three.js/wiki/Drawing-lines
+    var material, points = [], lineGeometry, limit = 0, path;
+
+    _trajectory = trajectory.representativeInterpolatedCoordinatesAtIndex(currentFrame);
+    if (_trajectory === null || _trajectory.length == 0)
+      return null;
+
+    material = new THREE.MeshPhongMaterial({color: color});
+    material.matrixAutoUpdate = true;
```
antgonza: Just to confirm, do these parameters need to be set after creation? In other words, can these be `material = new THREE.MeshPhongMaterial({color: color, matrixAutoUpdate: true, transparent: false});`. I checked the documentation and couldn't confirm ...

emperor/support_files/js/draw.js
```diff
@@ -295,11 +299,80 @@ define(['underscore', 'three', 'jquery'], function(_, THREE, $) {
...
+    //Tubes are straight segments, but adding vertices along them might change lighting effects
+    //under certain models and lighting conditions.
+    var tubeBufferGeom = new THREE.TubeBufferGeometry(path, (allPoints.length - 1) * NUM_TUBE_SEGMENTS, radius, NUM_TUBE_CROSS_SECTION_POINTS, false)
+
+    return new THREE.Mesh(tubeBufferGeom, material);
+  }
+  
+  function updateStaticTrajectoryDrawRange(trajectory, currentFrame, threeMesh)
+  {
+    //Blah, reverse engineering the number of points in a THREE tube is not fun, and may be implementation/version dependent.
```
antgonza: Blah?

ElDeveloper:
> @ElDeveloper; do you think it will be worth adding a "clear" example of how this new implementation improves animations?

Yes, although the only *real* dataset where this is showcased is a private one. I'll message you some data.

emperor/support_files/js/draw.js
dhakim87: As far as I can tell in the current implementation, the difference for boolean flags is only stylistic, but it heavily depends on the data types of the values of the map.

The map passed to the THREE.MeshPhongMaterial is named "parameters" in the MeshPhongMaterial constructor.  The final line of this constructor calls a function of Material called "setValues".  This setValues call loops over keys, warns about keys that it knows are deprecated, and defensively copies certain data types.  It's thus probably more future proof to use their setValues mechanism.  I'll make the change here, but I think updating existing code for consistency with this pattern is out of the scope of this PR.

emperor/support_files/js/draw.js
dhakim87: Is there a preferred comment to mark functions we think most likely to break?  
I personally tend to run with TODO FIXME HACK.  I'll remove the Blah.

emperor/support_files/js/draw.js
dhakim87: Also, matrixAutoUpdate is not a member of THREE.MeshPhongMaterial, but of the Mesh object itself.  Unclear what was intended or if this property of meshes needs to be set (True by default?  When would this property make a difference?).  Since setValues alerts the developer to this, I've updated my methods to use the constructor as you suggested, and removed the matrixAutoUpdate: true.

dhakim87:
Can someone point me towards the js formatter script that the travis-ci wants me to run?  
--- ...Some of the errors reported by GJsLint may be auto-fixable using the script
**fixjsstyle**... ---

emperor/support_files/js/draw.js
ElDeveloper: FIXME should be good. Let's ensure we exercise this enough that we are confident this won't break easily 👍

ElDeveloper:
> Can someone point me towards the js formatter script that the travis-ci wants me to run?

Sure thing, you need to install [closure-linter](https://github.com/google/closure-linter). Note that this only works with py27 so you might need a different environment. Let me know if you run into any issues.

emperor/support_files/js/animations-controller.js
```diff
@@ -538,19 +544,35 @@ define([
     var radius = view.getGeometryFactor();
     radius *= 0.45 * this.getRadius();
 
-    view.tubes.forEach(function(tube) {
-      if (tube === undefined) {
+    var i = 0;
+    for (i = 0; i < this.director.trajectories.length; i++) {
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/trajectory.js
```diff
@@ -186,7 +186,34 @@ define([
...
+   */
+  TrajectoryOfSamples.prototype.representativeInterpolatedCoordinatesAtIndex = function(idx) {
+    if (idx === 0)
+      return null;
+    if (this.interpolatedCoordinates.length -1 <= idx)
+      return null;
+      
+    lastStaticPoint = this.coordinates[this._intervalValues[idx]];
+    interpPoint = this.interpolatedCoordinates[idx];
+    if (lastStaticPoint.x === interpPoint.x
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/animations-controller.js
dhakim87: Done

emperor/support_files/js/draw.js
dhakim87: Done

emperor/support_files/js/animate.js
```diff
@@ -137,6 +137,15 @@ function(_, trajectory) {
      * @default -1
      */
     this.currentFrame = -1;
+    
+    /*
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/draw.js
dhakim87: Done for the new methods I've added.  Should discuss using this convention separately in the rest of the codebase as a separate item.

emperor/support_files/js/draw.js
dhakim87: Duplicate of ElDeveloper's comments on draw.js above.  Collapsing conversation.  Target should be to make the updateStaticTrajectoryDrawRange function as robust as possible.

emperor/support_files/js/draw.js
```diff
@@ -275,32 +279,111 @@ define(['underscore', 'three', 'jquery'], function(_, THREE, $) {
...
+                                    NUM_TUBE_CROSS_SECTION_POINTS,
+                                    false);
 
     return new THREE.Mesh(lineGeometry, material);
   }
 
+  /**
+   * Disposes a trajectory line dynamic mesh
+   */
+  function disposeTrajectoryLineDynamic(mesh) {
```
ElDeveloper: Both of the dispose functions do exactly the same thing, perhaps worth merging them into `disposeTrajectoryLine`?

dhakim87: Yeah, I was torn on this too.  I split them because the underlying geometry and material objects inside of the mesh are what needs to be disposed, and those are both shareable across mesh instances.  I was leaning towards a pattern where we put dispose (destructor) next to draw (constructor), so that changes in one function can be easily mirrored to the other.  If we have no short term plan to add geometry or material reuse and want to add as a rule instead that these are copied in every mesh, I think we should use a single function for disposing arbitrary meshes, which I think is just as good.  

On a side note, we're technically leaking Axis meshes when switching axes, so I figured a dispose method per type of scene object made sense when I realized I didn't want to trace the code to find out how axis meshes were initialized.

wasade:
BTW, I haven't been able to recreate the locking up I observed that I had mentioned in email, so from my perspective I'm :+1:

ElDeveloper:
Thanks so much @dhakim87 this is going to be hella useful!

PR #727, opened 2019-08-02
Parallel Plots

dhakim87:
This branch prototypes the ability to switch between scatter and parallel plots in the Axes tab.  Parallel Plot display is a GL_LINES mesh that is computed once and drawn with shaders mirroring the _fast_init call.  The Axes controller is setting a viewType field on the view that then modifies how the scene is rendered.  

There are a couple of noteworthy changes that we might want to discuss further:
When swapping scatter and parallel plots, we remove the existing meshes from the scene and add new ones.  This can get weird in terms of maintaining controller properties (color, scale etc...) and has some overhead for tearing down and creating the new meshes.  I'm also using a couple of dispose calls directly from sceneplotview3d since we don't have a better framework in place for managing gl objects.  

Parallel plots require just slightly more information than scatters for normalization purposes, the same dimension ranges that are required to draw axes.  I moved the computation of the dimension ranges into a new model object to shift it out of the sceneplotview3d.  

For (re)applying controller properties, we're just using a fromJson(toJson) serialization trick, and that seems to work well for small datasets, but something isn't scaling well at 20000+.  Since all controllers get reapplied at the same time, there's also potentially overhead in state changes that get overridden, and the potential for inconsistent intermediate states while the controllers are being applied.


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -663,6 +597,47 @@ define([
...
+        if (view.viewType == 'parallel-plot') {
+          for (i = 0; i < lines.length; i++)
+            scope.scene.remove(lines[i]);
+          for (i = 0; i < ellipsoids.length; i++)
+            scope.scene.remove(ellipsoids[i]);
+        }
+        if (view.viewType == 'scatter') {
+          for (i = 0; i < lines.length; i++)
+            scope.scene.add(lines[i]);
+          for (i = 0; i < lines.length; i++)
```
ElDeveloper: Is this predicate right, or should it be ellipsoids.length?

emperor/support_files/js/view.js
```diff
@@ -67,11 +80,23 @@ function DecompositionView(decomp, asPointCloud) {
...
    */
   this.tubes = [];
+
   /**
    * Array of THREE.Mesh objects on screen (represent samples).
    * @type {THREE.Mesh[]}
    */
   this.markers = [];
+
+  /**
```
ElDeveloper: Would you mind including the `@type` documentation for these two properties?

emperor/support_files/js/view.js
```diff
@@ -93,20 +118,30 @@ function DecompositionView(decomp, asPointCloud) {
...
+  this._initGeometry();
+}
+
+DecompositionView.prototype._initGeometry = function() {
+  this.oldMarkers = this.markers;
+  if (this.oldMarkers.length > 0)
+    this.needsSwapMarkers = true;
+  this.markers = [];
+
+  //TODO FIXME HACK:  Do we need to swap lines as well?
```
ElDeveloper: Can you elaborate on what you mean in this comment?

emperor/support_files/js/view.js
```diff
@@ -93,20 +118,30 @@ function DecompositionView(decomp, asPointCloud) {
...
   }
-
-  /**
-   * True when changes have occured that require re-rendering of the canvas
-   * @type {boolean}
-   */
   this.needsUpdate = true;
-}
+};
+
```
ElDeveloper: Can you delete this line?

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+ * Parallel plots closely mirroring the shader enabled _fastInit calls
+ */
+DecompositionView.prototype._fastInitParallelPlot = function()
+{
+  var positions, colors, scales, opacities, visibilities, geometry, cloud;
+
+  // We're really just drawing a bunch of line strips...
+  // highly doubt shaders are necessary for this...
+  var vertexShader = `
+    attribute float scale;
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+    varying float vOpacity;
+    varying float vVisible;
+    
+    void main() {
+      vColor = color;
+      vOpacity = opacity;
+      vVisible = visible;
+    
+      gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+      gl_PointSize = 2.0;
```
ElDeveloper: I think we don't need this, right?

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+    `;
+
+  var fragmentShader = `
+    precision mediump float;
+    varying vec3 vColor;
+    varying float vOpacity;
+    varying float vVisible;
+    
+    void main() {
+      if (vVisible <= 0.0)
```
ElDeveloper: I think this should be  `(vVisible <= 0.0 || vOpacity <= 0)`.

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+
+  var allDimensions = _.range(this.decomp.dimensions);
+
+  // We'll build the line strips as GL_LINES for simplicity, at least for now,
+  // by doubling up vertex positions at each of the intermediate axes.
+  var numPoints = (allDimensions.length * 2 - 2) * (this.decomp.length);
+
+  positions = new Float32Array(numPoints * 3);
+  colors = new Float32Array(numPoints * 3);
+  scales = new Float32Array(numPoints);
```
ElDeveloper: Can you remove this line.

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+  var material = new THREE.ShaderMaterial({
+    vertexShader: vertexShader,
+    fragmentShader: fragmentShader,
+    transparent: true
+  });
+
+  geometry = new THREE.BufferGeometry();
+  geometry.addAttribute('position', new THREE.BufferAttribute(positions, 3));
+  geometry.addAttribute('color', new THREE.BufferAttribute(colors, 3));
+  geometry.addAttribute('scale', new THREE.BufferAttribute(scales, 1));
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+  geometry.addAttribute('opacity', new THREE.BufferAttribute(opacities, 1));
+  geometry.addAttribute('visible', new THREE.BufferAttribute(visibilities, 1));
+
+  lines = new THREE.LineSegments(geometry, material);
+
```
ElDeveloper: Can you remove this line?

emperor/support_files/js/view.js
```diff
@@ -337,6 +372,125 @@ DecompositionView.prototype._fastInit = function() {
...
+      var interpVal = (plottable.coordinates[j] - globalMin) / (maxMinusMin);
+      geometry.attributes.position.setXYZ(attributeIndex,
+                                        j,
+                                        interpVal,
+                                        0);
+
+      geometry.attributes.color.setXYZ(attributeIndex, 1, 0, 0);
+      geometry.attributes.visible.setX(attributeIndex, 1);
+      geometry.attributes.opacity.setX(attributeIndex, 1);
+      geometry.attributes.scale.setX(attributeIndex, 1);
```
ElDeveloper: This will need to be deleted.

emperor/support_files/js/view.js
```diff
@@ -395,6 +556,9 @@ DecompositionView.prototype.updatePositions = function() {
     });
     cloud.geometry.attributes.position.needsUpdate = true;
   }
+  else if (this.decomp.isScatterType() && this.viewType === 'parallel-plot') {
+    //TODO:  Do we need to do anything when axes are changed in parallel plots?
```
ElDeveloper: No, in fact this shouldn't happen at all.

emperor/support_files/js/multi-model.js
```diff
@@ -0,0 +1,103 @@
...
+//    /**
+//     * @class MultiModel
+//     *
+//     * A model object containing metrics merged from multiple underlying
+//     * DecompositionModel objects that share the same coordinate space.
+//     *
+//     * The multiModel contains min and max range of all model objects in the
+//     * scene, allowing you to draw axes or normalize data for a parallel plot.
+//     */
+  function MultiModel(underlyingModels) {
```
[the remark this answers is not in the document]
dhakim87: Yes, I had to move the union ranges code out of the sceneplotview3d object and into a model class so that I could use it from the view classes.  I hate the name multi-model though.

emperor/support_files/js/view.js
dhakim87: I don't have a biplot dataset (that's where lines come from right?), so I'm not sure what the behavior is or should be when there are lines in the scene.  I think I went with disabling lines and re-enabling lines, but presumably the parallel plot view could have lines in it at some point that would require new swapping logic.

emperor/support_files/js/view.js
dhakim87: PointSize?  Probably not.

emperor/support_files/js/axes-controller.js
```diff
@@ -33,6 +33,32 @@ define([
...
+    this.$viewTypeDiv.attr('title', 'Change the selected View Type');
+    this.$radioScatter = $('<input type="radio" name="emperor.viewType"' +
+      'value="scatter"> Scatter </input>');
+    this.$radioParallelPlot = $('<input type="radio" name="emperor.viewType"' +
+      'value="parallel-plot"> Parallel Plot </input>');
+
+    this.$viewTypeDiv.append(this.$radioScatter);
+    this.$viewTypeDiv.append(this.$radioParallelPlot);
+
+    this.$radioScatter.change(function() {
```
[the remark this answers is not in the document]
dhakim87: I disabled buildDisplayTable when parallel plot mode is on.  That hides the three axis visibility and invert picker.

emperor/support_files/js/axes-controller.js
```diff
@@ -33,6 +33,32 @@ define([
...
+    });
+
+    this.$radioParallelPlot.change(function() {
+      decompViewDict['scatter'].setViewType('parallel-plot');
+    });
+
+    this.$header.prepend($('<hr>'));
+    this.$header.prepend(this.$viewTypeDiv);
+
+
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/controller.js
```diff
@@ -17,12 +17,14 @@ define([
     'svgrenderer',
     'draw',
     'canvasrenderer',
-    'canvastoblob'
+    'canvastoblob',
+    'multi-model',
+//    'view-types',
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/multi-model.js
```diff
@@ -0,0 +1,103 @@
+define(['underscore'],
+function(_) {
+//    /**
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Uhh... maybe done?  The line right after scale?

emperor/support_files/js/view.js
dhakim87: Removed.  Done.

emperor/support_files/js/view.js
dhakim87: Updated

emperor/support_files/js/view.js
dhakim87: 422?  I guess so?  I'm starting to wonder if your editor is showing something that I'm not seeing though.

ElDeveloper: Not sure. Here I was referring to the "scales" attribute given that we can't really change the size of the lines, right?

emperor/support_files/js/view.js
dhakim87: Ohhh!  The scale lines!  Done.

emperor/support_files/js/view.js
dhakim87: Ahh, the scale lines, done.

emperor/support_files/js/view.js
dhakim87: Done.

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Removed scale

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: We hit this line as a result of the controller save to json/load from json hack when switching between scatter and parallel plot.  I do agree that doing nothing here is a bit of a bandaid, but we need to add change tracking (like the UIState properties) to all the saved/loaded properties in order to avoid this.  

Stack that hits this line:
console.trace() view.js:556:13
    updatePositions file:///Users/Dan/knightLab/emperor/emperor/support_files//js/view.js:556
    changeVisibleDimensions file:///Users/Dan/knightLab/emperor/emperor/support_files//js/view.js:689
    fromJSON file:///Users/Dan/knightLab/emperor/emperor/support_files//js/axes-controller.js:760
    forceRefresh file:///Users/Dan/knightLab/emperor/emperor/support_files//js/abc-view-controller.js:192
    render file:///Users/Dan/knightLab/emperor/emperor/support_files//js/controller.js:506
    each jQuery
    render file:///Users/Dan/knightLab/emperor/emperor/support_files//js/controller.js:500
    animate file:///Users/Dan/knightLab/emperor/new-emperor.html:174

emperor/support_files/js/shape-controller.js
```diff
@@ -115,6 +115,10 @@ define([
    */
   ShapeController.prototype.setPlottableAttributes =
       function(scope, shape, group) {
+
```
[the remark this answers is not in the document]
dhakim87: The tab is now hidden when switching viewTypes.  That is not sufficient to remove this line as yet: setPlottableAttributes will be called when switching viewTypes due to the controller save to json/load from json hack.  Should I be moving the controller save load logic into something with change tracking as part of this PR?

emperor/support_files/js/animations-controller.js
```diff
@@ -547,14 +547,14 @@ define([
...
 
-    view.tubes = this.director.trajectories.map(function(trajectory) {
-      color = scope._colors[trajectory.metadataCategoryName] || 'red';
-
-      var tube = drawTrajectoryLine(trajectory, scope.director.currentFrame,
-                                    color, radius);
-      return tube;
-    });
-
+    if (view.viewType !== 'parallel-plot') {
```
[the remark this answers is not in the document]
dhakim87: Same status as ShapeController above.

dhakim87:
Splitting changes from your review into separate conversations.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
     throw Error('Not implemented');
   };
 
+  /**
```
ElDeveloper: - When switching to the parallel plot view, it would be great if the camera could try to fit most of the plot or at least 5 dimensions of the plot on screen.
    - Simiarly, when clicking "recenter camera" it should go to that view that fits ~5 dimensions.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
...
+   * Writes the current settings to the active decomposition view(s).
```
ElDeveloper: - Looks like save to PNG is not working with parallel plots.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
...
+   * Will be called when the decomposition view is swapped out
```
ElDeveloper: - Once parallel plots are enabled. There's a few things that need to change in the UI:

    - the "axes dropdowns" in the Axes tab should be disabled.

    - The Scale, shape and animations tab should be all disabled or hidden.

    - In the context menu:
        - The export as SVG and export to VEGA should also be disabled.

        - The option to toggle auto-rotation should also be disabled.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
...
+   * for a different view type.
```
ElDeveloper: - Ray-casting is not working when looking at a paralle plot i.e. the lines won't reveal their name when you click on them.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
...
+   */
```
ElDeveloper: - It looks like there's a few unit tests that need to be fixed or moved over. Let me know if you would like help with this.
    - Namely ScenePlotView3D, MultiModel and DecompositionView will need a fair amount of work.

emperor/support_files/js/abc-view-controller.js
```diff
@@ -183,5 +183,14 @@ define([
...
+  EmperorViewControllerABC.prototype.forceRefresh = function() {
```
ElDeveloper: - With regards to the problem of re-applying settings being a slow operation, perhaps we can instead go marker by marker and copy-over the attributes. Instead of updating the UI and triggering a number of unnecessary calls.

emperor/support_files/js/abc-view-controller.js
dhakim87: Done.  Presumably CanvasRenderer can't handle shaders, so I set it to use the other renderer, same as you did for large point cloud datasets.

dhakim87: NOTE:  I have no way to test the svg mode.  I can't even get svg mode to work correctly on scatter plots on my machine.

emperor/support_files/js/abc-view-controller.js
dhakim87: Done.  Also set auto-rotation to false when switching to parallel-plot mode.

emperor/support_files/js/abc-view-controller.js
dhakim87: Added this.  It can't be pixel perfect: all calculations are done with a raycaster precision value in model coordinates, but if the user zooms in, they'll click the correct line every time.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -263,6 +279,59 @@ define([
...
+    }
+    else if (viewType === 'parallel-plot')
+    {
+      var w = this.decModels.dimensionRanges.max.length;
+
+      // Set up the camera
+      camera = new THREE.OrthographicCamera(0, w, 1, 0);
+      camera.position.set(0, 0, 1); //Must set positive Z because near > 0
+      camera.zoom = 0.7;
+    }
```
ElDeveloper: Can you add an `else` clause and raise an error with the name of the viewType?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -263,6 +279,59 @@ define([
...
+
+  /**
+   * Builds a camera controller (for scatter or parallel plot)
+   */
+  ScenePlotView3D.prototype.buildCamController = function(viewType, cam, view) {
+    /**
+     * Object used to interact with the scene. By default it uses the mouse.
+     * @type {THREE.OrbitControls}
+     */
+    var control = new THREE.OrbitControls(cam,
```
ElDeveloper: Can you join this and the next line? Doesn't look like these will exceed the 80 column limit.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -81,72 +86,83 @@ define([
...
-    var max = _.max(decViews.scatter.decomp.dimensionRanges.max);
-    var frontFrust = _.min([max * 0.001, 1]);
-    var backFrust = _.max([max * 100, 100]);
-
-    /**
-     * Camera used to display the scene.
-     * @type {THREE.PerspectiveCamera}
-     */
-    // these are placeholders that are later updated in updateCameraAspectRatio
-    this.camera = new THREE.OrthographicCamera(-50, 50, 50, -50);
```
ElDeveloper: Would you mind re-adding this property? It looks like it is still in use but its underlying value will change depending on the view-type. Perhaps leave this as:

```javascript
this.camera = null;
```

And also make a note that this changes depending on what view is active. Seems to me that the `scatterCam` and the `paralleCam` could both be private properties?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -81,72 +86,83 @@ define([
...
     this.light = new THREE.DirectionalLight(0x999999, 2);
     this.light.position.set(1, 1, 1).normalize();
-    this.camera.add(this.light);
+    this.scatterCam.add(this.light);
+
+    /**
+     * Camera used to display the parallel plot scene.
+     * @type {THREE.OrthographicCamera}
+     */
+    this.parallelCam = this.buildCamera('parallel-plot');
```
ElDeveloper: Would you mind explaining the rationale for having two cameras instead of having one that changes some of its attributes depending on the current UI state?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -299,14 +368,29 @@ define([
...
 
+    // if a decomposition uses a point cloud, or
+    // if a decomposition uses a parallel plot,
+    // update the default tolerance as
+    // it is otherwise too large and error-prone
+    var scope = this;
+    var updateRaycasterLinePrecision = function(evt) {
+      if (UIState.getProperty('view.viewType') === 'parallel-plot')
+        scope._raycaster.linePrecision = 0.01;
+      else
```
ElDeveloper: Do we need to set a line precision if the view type is not a parallelPlot?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -299,14 +368,29 @@ define([
...
+        scope._raycaster.linePrecision = 1;
+    };
+    var updateRaycasterPointPrecision = function(evt) {
+      if (UIState.getProperty('view.usesPointCloud'))
+        scope._raycaster.params.Points.threshold = 0.01;
+      else
+        scope._raycaster.params.Points.threshold = 1;
```
ElDeveloper: Same as above.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -518,20 +532,57 @@ define([
...
+
+        //Assuming a camera zoom of 1:
+        var viewWUnits = cam.right - cam.left;
+        var viewHUnits = cam.top - cam.bottom;
+
+        //These are world sizes of label for a camera zoom of 1
+        var labelWUnits = labelWPix * viewWUnits / viewWPix;
+        var labelHUnits = labelHPix * viewHUnits / viewHPix;
+
+        //TODO FIXME HACK:  Note that our options here are to scale each
```
ElDeveloper: Thanks for pointing this out. Can you remove the `TODO FIXME HACK` tag? This seems like a reasonable note to leave though 👍 and I agree with the choice.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -631,28 +700,44 @@ define([
...
+      this.control.target.set(0, 0, 1); //Must set positive Z because near > 0
+      this.camera.position.set(0, 0, 1); //Must set positive Z because near > 0
+      this.camera.updateProjectionMatrix();
+      this.updateCameraAspectRatio();
+      this.control.saveState();
+      this.needsUpdate = true;
+    }
   };
 
+  ScenePlotView3D.prototype.NEEDS_RENDER = 1;
```
ElDeveloper: Is there any way we could do without these properties?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -663,6 +748,47 @@ define([
...
+    //tree.
+    var anyMarkersSwapped = false;
+    _.each(this.decViews, function(view) {
+      if (view.needsSwapMarkers) {
+        anyMarkersSwapped = true;
+
+        var oldMarkers = view.getAndClearOldMarkers();
+        for (var i = 0; i < oldMarkers.length; i++) {
+          scope.scene.remove(oldMarkers[i]);
+          oldMarkers[i].material.dispose(); //FIXME:  What is our plan for this?
```
ElDeveloper: Once disposed we should be good, or am I missing something?

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -723,9 +849,15 @@ define([
       this.drawAxesLabelsWithColor(this.axesColor);
     }
 
+    var retVal = 0;
+    if (anyMarkersSwapped)
+      retVal |= ScenePlotView3D.prototype.NEEDS_CONTROLLER_REFRESH;
```
ElDeveloper: Any rationale for having this be a bitwise OR?

emperor/support_files/js/view.js
ElDeveloper: Yep that should be the right answer. Can you remove this comment line then?

emperor/support_files/js/view.js
```diff
@@ -353,6 +378,130 @@ DecompositionView.prototype._fastInit = function() {
...
 };
 
+/**
+ * Parallel plots closely mirroring the shader enabled _fastInit calls
+ */
+DecompositionView.prototype._fastInitParallelPlot = function()
+{
+  var positions, colors, opacities, visibilities, geometry, cloud;
+
+  // We're really just drawing a bunch of line strips...
```
ElDeveloper: Can you remove these two lines? It is my understanding that we definitely need shaders, right? Otherwise the additional required overhead with THREE.js objects would make large plots way too slow.

emperor/support_files/js/view.js
```diff
@@ -353,6 +378,130 @@ DecompositionView.prototype._fastInit = function() {
...
+    // positions due to the use of lines rather than line strips.
+    for (var j = 0; j < allDimensions.length; j++)
+    {
+      //normalize by global range bounds
+      var globalMin = this.allModels.dimensionRanges.min[allDimensions[j]];
+      var globalMax = this.allModels.dimensionRanges.max[allDimensions[j]];
+      var maxMinusMin = globalMax - globalMin;
+      var interpVal = (plottable.coordinates[j] - globalMin) / (maxMinusMin);
+      geometry.attributes.position.setXYZ(attributeIndex,
+                                        j,
```
ElDeveloper: Any chance you could align the indentation of these arguments? And the ones in line 473?

emperor/support_files/js/view.js
```diff
@@ -353,6 +378,130 @@ DecompositionView.prototype._fastInit = function() {
...
+                                        interpVal,
+                                        0);
+
+      geometry.attributes.color.setXYZ(attributeIndex, 1, 0, 0);
+      geometry.attributes.visible.setX(attributeIndex, 1);
+      geometry.attributes.opacity.setX(attributeIndex, 1);
+      attributeIndex++;
+
+      if (j == 0 || j == allDimensions.length - 1)
```
ElDeveloper: Can you add a comment for why we skip the first and last elements?

emperor/support_files/js/view.js
ElDeveloper: Got it, this seems fine then.

emperor/support_files/js/view.js
```diff
@@ -694,6 +856,19 @@ DecompositionView.prototype.setColor = function(color, group) {
     });
     cloud.geometry.attributes.color.needsUpdate = true;
   }
+  else if (UIState['view.viewType'] == 'parallel-plot' &&
+           this.decomp.isScatterType()) {
+    var lines = this.markers[0];
+    color = new THREE.Color(color);
+    var numPoints = (this.decomp.dimensions * 2 - 2);
+    group.forEach(function(plottable) {
+    var startIndex = plottable.idx * numPoints;
```
ElDeveloper: Would you mind indenting the code in this function?

emperor/support_files/js/view.js
```diff
@@ -694,6 +856,19 @@ DecompositionView.prototype.setColor = function(color, group) {
...
+    var endIndex = (plottable.idx + 1) * numPoints;
+    for (var i = startIndex; i < endIndex; i++)
+        lines.geometry.attributes.color.setXYZ(i, color.r, color.g, color.b);
```
ElDeveloper: This line looks over indented, can you remove two of the leading spaces?

emperor/support_files/js/view.js
```diff
@@ -986,6 +1192,14 @@ DecompositionView.prototype._buildVegaSpec = function() {
   };
 };
 
+
+DecompositionView.prototype.getAndClearOldMarkers = function() {
```
ElDeveloper: Can you add a docstring?

emperor/support_files/js/axes-controller.js
```diff
@@ -33,6 +34,46 @@ define([
     EmperorViewController.call(this, container, title, helpmenu,
                                decompViewDict);
 
+    this.$viewTypeDiv = $('<div name="emperor-viewtype-div"></div>');
+    this.$viewTypeDiv.css('margin', '0 auto');
```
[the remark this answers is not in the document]
dhakim87: Done.

emperor/support_files/js/axes-controller.js
```diff
@@ -33,6 +34,46 @@ define([
...
+    this.$viewTypeDiv.css('width', '100%');
+    this.$viewTypeDiv.css({
+        'margin': '0 auto',
+        'width': '100%',
+        'height': '100%'
+    });
+    this.$viewTypeDiv.attr('title', 'Change the selected View Type');
+
+    if (UIState['view.viewType'] === 'scatter')
```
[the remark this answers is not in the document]
dhakim87: Done.

emperor/support_files/js/axes-controller.js
```diff
@@ -196,6 +237,12 @@ define([
     }
 
     var view = this.getView(), scope = this;
+
+    if (UIState['view.viewType'] === 'parallel-plot') {
```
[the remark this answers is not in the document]
dhakim87: 239 refers to the view and scope declarations?  Sure, there is no problem moving the fail out above these declarations.  

Done.

emperor/support_files/js/multi-model.js
```diff
@@ -0,0 +1,112 @@
+define(['underscore'],
+function(_) {
+  /**
+   * @class MultiModel
+   *
+   * A model object containing metrics merged from multiple underlying
+   * DecompositionModel objects that share the same coordinate space.
+   *
+   * The multiModel contains min and max range of all model objects in the
```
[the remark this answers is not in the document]
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Done

emperor/support_files/js/view.js
dhakim87: Done

dhakim87:
We had a conversation on properly hooking up UIState with require.js and I don't want those thoughts to disappear:

The unit test framework runs multiple tests within the same instance of the application, and sharing a UIState between them complicates the unit test procedure.  We should think about mechanisms to pass a separate UIState with each test, either by swapping the UIState for a mock object or by configuring require.js in some special way with each test.  

The jupyter notebook configuration likely shares UIState objects between multiple instances within a single notebook.  This needs to be handled appropriately before the PR goes in.

ElDeveloper:
> The jupyter notebook configuration likely shares UIState objects between multiple instances within a single notebook. This needs to be handled appropriately before the PR goes in.

Thanks for bringing this up, let's talk about this and see what can we do right now. It sounds like this will need to be a require-specific solution that we'll need to think about.

antgonza:
@dhakim87, this is pretty nice; thank you for your help with this.

A couple of minor suggestions and a question:
- if a user clicks on the name Scatter/Parallel-Plot the selection should also change; currently you have to click on the circle/selector
- should the file name of the PNG (high resolution) be emperor_parallel_plots.png so the user can quickly differentiate between the regular and the parallel plots?
- what's the plan for the labels?

emperor/support_files/js/axes-controller.js
```diff
@@ -24,15 +25,61 @@ define([
...
 
+    this.$viewTypeDiv = $('<div name="emperor-viewtype-div"></div>');
+    this.$viewTypeDiv.css({
+        'margin': '0 auto',
+        'width': '100%',
+        'height': '100%'
+    });
+    this.$viewTypeDiv.attr('title', 'Change the selected View Type');
+
+    var radioName = "emperor.viewType_" + this.identifier;
```
ElDeveloper:
```suggestion
    var radioName = 'emperor.viewType_' + this.identifier;
```

ElDeveloper:
Thanks so much @dhakim87!!! This is going to be awesome!

PR #728, opened 2019-08-12
Fix a problem where invisible or transparent objects would be intersected

ElDeveloper:
This PR explicitly removes objects that are not visible when raycasting.


ElDeveloper:
> Code looks good, just wondering if it this new code news new tests ...

Just went through the tests for ScenePlotView and it doesn't look like we were ever able to test point intersections before. I vaguely remember this being related to the fact we use an SVG renderer for testing. For this specific case we need to render shaders which the SVG renderer isn't even capable of handling at the moment. So this PR should be good to go as is.

🙏

ElDeveloper:
Thanks @antgonza!

Issue #730, opened 2019-08-23
ignore-missing-samples doesn't apply to arrows


mortonjt:
It would be very convenient if this could also apply to arrows.  You can imagine scenarios where some features are just not annotated.

PR #731, opened 2019-08-23
Enabling filtering of missing features

mortonjt:
Addresses #730 

Now when the ignore_missing_samples flag is enabled, features can also be enabled.

This is an easy fix - but maybe the terminology will need to be readdressed (i.e. change ignore_missing_samples -> ignore_missing).

### DIFF
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -250,7 +250,7 @@ class Emperor(object):
             self.feature_mf = \
                 self._validate_metadata(feature_mapping_file,
                                         self.ordination.features,
-                                        ignore_missing_samples=False)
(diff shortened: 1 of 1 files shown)


mortonjt:
@ElDeveloper , is there any chance that this can get merged in?  This is becoming a bottleneck when mass producing emperor plots.

mortonjt:
@ElDeveloper , I think you are right - this block needs more investigation.  Just ran into another issue when I use the patch that I proposed. See #733

ElDeveloper:
I've opened a new PR based on these changes, see #734. . It fixes #730 and #733.

Issue #733, opened 2019-09-05
Possible issue when validating feature metadata


mortonjt:
I'm running into an issue when there is missing data in the feature metadata

```bash
qiime emperor biplot   --i-biplot biplot.qza   --m-sample-metadata-file 2019-08-09_doty_metabolomics_normalized_cpm_feature_md.tsv metab_diffs.qza   --m-feature-metadata-file Doty_taxonomy.tsv micro_diffs.qza   --p-number-of-features 100   --p-ignore-missing-samples   --o-visualization omics-biplot_.qzv --verbose
```

```
Traceback (most recent call last):
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/q2cli/commands.py", line 327, in __call__
    results = action(**arguments)
  File "</Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/decorator.py:decorator-gen-388>", line 2, in biplot
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/qiime2/sdk/action.py", line 240, in bound_callable
    output_types, provenance)
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/qiime2/sdk/action.py", line 445, in _callable_executor_
    ret_val = self._callable(output_dir=temp_dir, **view_args)
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/q2_emperor/_plot.py", line 95, in biplot
    plot_name='biplot')
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/q2_emperor/_plot.py", line 41, in generic_plot
    procrustes=procrustes, remote='.')
  File "/Users/jmorton/Documents/dev/emperor/emperor/core.py", line 253, in __init__
    ignore_missing_samples)
  File "/Users/jmorton/Documents/dev/emperor/emperor/core.py", line 339, in _validate_metadata
    data=data)
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/pandas/core/frame.py", line 424, in __init__
    copy=copy)
  File "/Users/jmorton/miniconda3/envs/qiime2-2019.7/lib/python3.6/site-packages/pandas/core/internals/construction.py", line 167, in init_ndarray
[...]

PR #734, opened 2019-09-11
ENH:Allow for missing features in feature metadata

### LINKED ISSUE #730: ignore-missing-samples doesn't apply to arrows

ElDeveloper:
Allows for feature metadata to not include some of the features
represented in the ordination. The errors and warnings have been updated
to properly refer to either features or samples. I've added a couple of
tests to validate the error/warning messages.

Fixes #730  and #733

### DIFF
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -325,11 +325,22 @@ class Emperor(object):
                              ' to the same dataset.' % kind)
 
         if difference and not ignore_missing_samples:
-            elements = ', '.join(sorted([str(i) for i in difference]))
+            # sort the elements so we have a deterministic output
+            difference = sorted([str(i) for i in difference])
+
+            # if there's more than 5 missing elements, truncate the list
+            if len(difference) > 5:
+                elements = ', '.join(difference[:5])
+                suffix = ("Showing only the first 5 %ss out of %d: %s ..." %
+                          (kind, len(difference), elements))
+            else:
+                elements = ', '.join(difference)
+                suffix = ("Offending %ss: %s" % (kind, elements))
+
             raise KeyError("There are %ss not included in the %s mapping "
                            "file. Override this error by using the "
-                           "`ignore_missing_samples` argument. Offending "
-                           "%ss: %s" % (kind, kind, kind, elements))
+                           "`ignore_missing_samples` argument. %s" %
+                           (kind, kind, suffix))
         elif difference and ignore_missing_samples:
             warnings.warn("%d out of %d %ss have no metadata and are being"
                           " included with a placeholder value." %
--- a/emperor/core.py
+++ b/emperor/core.py
@@ -96,7 +96,7 @@ class Emperor(object):
         identifiers as the identifiers in ``ordination``.
     ignore_missing_samples: bool, optional
         If set to `True` samples and features without metadata are included by
-        setting all metadata values to: ``This element has not metadata``. By
+        setting all metadata values to: ``This element has no metadata``. By
         default an exception will be raised if missing elements are
         encountered. Note, this flag only takes effect if there's at least one
         overlapping element.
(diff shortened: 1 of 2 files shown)


emperor/core.py
```diff
@@ -250,7 +251,7 @@ def __init__(self, ordination, mapping_file, feature_mapping_file=None,
             self.feature_mf = \
                 self._validate_metadata(feature_mapping_file,
                                         self.ordination.features,
-                                        ignore_missing_samples=False)
+                                        ignore_missing_samples, kind='feature')
```
mortonjt: this is awesome!

ElDeveloper:
This PR is ready for review if anyone would like to have a look.

emperor/core.py
```diff
@@ -95,10 +95,11 @@ class Emperor(object):
         A list of the OrdinationResults objects with the same sample
         identifiers as the identifiers in ``ordination``.
     ignore_missing_samples: bool, optional
-        If set to `True` samples without metadata are included by setting all
-        metadata values to: ``This sample has not metadata``. By default an
-        exception will be raised if missing samples are encountered. Note, this
-        flag only takes effect if there's at least one overlapping sample.
+        If set to `True` samples and features without metadata are included by
+        setting all metadata values to: ``This element has not metadata``. By
```
wasade: What about "This element is missing metadata"?

emperor/core.py
```diff
@@ -302,39 +303,43 @@ def _repr_html_(self):
...
+        if difference == ordination_elements:
+            raise ValueError('None of the %s identifiers match between the'
                              ' metadata and the coordinates. Verify that you '
                              'are using metadata and coordinates corresponding'
-                             ' to the same dataset.')
+                             ' to the same dataset.' % kind)
 
         if difference and not ignore_missing_samples:
-            raise KeyError("There are samples not included in the mapping "
+            elements = ', '.join(sorted([str(i) for i in difference]))
```
wasade: Can this be a truncated list? If thousands of features are missing, the output will be substantial especially if they're sOTUs

emperor/core.py
wasade: ...also the actual text in the code was "This element has no metadata" which would also be okay. The "has not" was weird :)

emperor/core.py
```diff
@@ -302,39 +303,43 @@ def _repr_html_(self):
...
-                          (len(difference), len(ordination_samples)),
+                          (len(difference), len(ordination_elements), kind),
                           EmperorWarning)
 
-            # pad the missing samples
+            # pad the missing elements
             data = np.full((len(difference), metadata.shape[1]),
-                           'This sample has no metadata', dtype='<U27')
-            pad = pd.DataFrame(index=difference, columns=self.mf.columns,
+                           'This element has no metadata', dtype='<U28')
```
wasade: I *think* just using a dtype of str will allow numpy to determine the exact type it wants to use?

emperor/core.py
ElDeveloper: Got it, I'll correct the documentation then 👍

emperor/core.py
ElDeveloper: Good point, come to think about it I can do a fillna using the intermediate DataFrame, fixed - thanks!

emperor/core.py
ElDeveloper: I think it's ok since QIIME2 would hide the list of features by default and only if the user uses `--verbose` the full text will be displayed. What do you think?

wasade: Does it...? I ran q2-emperor from 2019.7 recently with malformed metadata and got a buffer full of output on the terminal :/

ElDeveloper: Got it, just addressed this 👍

wasade:
:+1:

ElDeveloper:
Thanks so much @wasade!

PR #740, opened 2019-10-26
Fix raycasting issue

ElDeveloper:
This bug was introduced when we resolved the conflicts after merging #727 and #728. Namely `intersect` was stored as global variable, the name match was used instead of marker, and the mesh objects were accessing non-existent properties.

@dhakim87 if you can have a look and merge that would be great!


dhakim87:
This branch has been merged into the parallel plots branch: dhakim87/new-api

ElDeveloper:
Thanks @dhakim87

PR #746, opened 2019-11-27
Fixes issue with slickgrid re-rendering

ElDeveloper:
This fixes an issue of category values not updating after a new metadata column is selected. The issue was a missing call to invalidate and render in the grid.

The animated GIF bellow shows the interface working again:

issue-744


antgonza:
Looks good, worth adding a test?

ElDeveloper:
Any ideas on how to test this? The data model is updated, it is only the 
view that doesn't get updated unless you force a refresh/render call.

antgonza:
Not sure if possible but perhaps checking the container and see that the new elements/values are in there? If not possible then is fine ...

ElDeveloper:
@antgonza, I just pushed some changes that add some tests. Albeit these are rather limited, I use jquery to see if we can find elements with the expected categories.

PR #749, opened 2020-01-09
ENH: Expand range for animation speed slider

ElDeveloper:
@fedarko, can you review this PR? It should help with the animations that you were putting together where transitions were slower than desired.


fedarko:
@ElDeveloper thanks! I installed your branch into my QIIME 2 environment, and on my system at least, it looks like increasing the slider speed above 10 causes the animation to just automatically complete for some reason. I've sent over the files to reproduce this problem.

PR #750, opened 2020-01-09
Fix _removeObjectsWithPrefix For Duplicate Object Names

### LINKED ISSUE #748: Duplicate axis created when using weird axis settings
This seems to me like a pretty obscure problem, but I think it is technically a bug.

## Steps to reproduce

1. Open up a plot or biplot in Emperor (I checked and this error is reproducible with normal scatterplots or biplots)

Screen Shot 2020-01-07 at 11 16 30 PM

2. Go to the `Axes` tab and set all axes to the same PC (I know, I know). As expected, all of the samples are constrained to a single line in the ordination space.

Screen Shot 2020-01-07 at 11 18 05 PM

3. Set any of the axes to a different PC. Now, for some reason, the axis display is duplicated:

Screen Shot 2020-01-07 at 11 20 16 PM

Note that the "direction" of the duplication (wherever the new axis display ends up in relation to the original one) seems to be based on the axis you changed in step 3 -- changing the first axis as shown in the screenshot shifts the duplicate axis horizontally, changing the second axis shifts the duplicate axis vertically (see below), and changing the third axis shifts the duplicate axis "towards the camera" (at least from the default camera settings).

Screen Shot 2020-01-07 at 11 22 50 PM

I don't know _why_ this error happens, but it might have ramifications for other parts of Emperor.

dhakim87:
Resolves #748 

The underlying issue is that axes were created with duplicate names in the scene when multiple axes were set to the same axis index.  This would cause objects to be leftover when switching states.  This was addressed by making _removeObjectsWithPrefix remove -all- objects with a given prefix, rather than 1 with each name.  This does however require one additional full search of the tree.

### DIFF
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -591,21 +591,26 @@ define([
 
   /**
    *
-   * Helper method to remove objects that match a prefix from the view's scene
-   * this method is used by removeAxes and removeAxesLabels. This function
-   * iterates "num" times, and for each iteration it finds and removes objects
-   * with the name of the form "prefix" + "iteration".
+   * Helper method to remove objects with some prefix from the view's scene
    *
-   * @param {String} prefix The label that will prepended to the iterating
-   * index.
+   * @param {String} prefix The prefix of object names to remove
    *
    */
   ScenePlotView3D.prototype._removeObjectsWithPrefix = function(prefix) {
     var scope = this;
-    _.each(_.range(this.decViews.scatter.decomp.dimensions), function(i) {
-      var axisLine = scope.scene.getObjectByName(prefix + i);
-      scope.scene.remove(axisLine);
-    });
+    var recursiveRemove = function(rootObj) {
+      if (rootObj.name != null && rootObj.name.startsWith(prefix)) {
+        scope.scene.remove(rootObj)
+      }
+      else {
+        // We can't iterate the children array while removing from it,
+        // So we make a shallow copy.
+        var childCopy = Array.from(rootObj.children)
+        for (child in childCopy)
+          recursiveRemove(childCopy[child])
+      }
+    }
+    recursiveRemove(this.scene);
   };
 
   /**


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -591,21 +591,26 @@ define([
...
+   * Helper method to remove objects with some prefix from the view's scene
    *
-   * @param {String} prefix The label that will prepended to the iterating
-   * index.
+   * @param {String} prefix The prefix of object names to remove
    *
    */
   ScenePlotView3D.prototype._removeObjectsWithPrefix = function(prefix) {
     var scope = this;
-    _.each(_.range(this.decViews.scatter.decomp.dimensions), function(i) {
```
ElDeveloper: After looking into this, I think the old function would still work, as long as we update the objects we iterate over. Instead of iterating over decomp.dimensions (which would go from 1 - N), we need to iterate over scenePlotView.visibleDimensions (which tracks the index of the visible dimensions). Something like this should work:


```javascript
  ScenePlotView3D.prototype._removeObjectsWithPrefix = function(prefix) {
    var scope = this;
    _.each(this.visibleDimensions, function(i) {
      var objectWithPrefix = scope.scene.getObjectByName(prefix + i);
      scope.scene.remove(objectWithPrefix);
    });
  };
```

dhakim87:
Either should work, as long as you can guarantee we generate separate remove calls for each object with a duplicate name.  If you’re worried about the cost of traversing the tree, we should double check that three.js doesn’t look up names with a full scan also.

ElDeveloper:
> If you’re worried about the cost of traversing the tree, we should double check that three.js doesn’t look up names with a full scan also.

This is a good point, as far as I recall they both follow a similar approach i.e. traversing the whole tree:
https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L352

Do you think we could do the removal of objects in a smarter way with this in mind?

Also, I'm looking into solving the issues with the build failing in #752.

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -591,21 +591,26 @@ define([
...
-      var axisLine = scope.scene.getObjectByName(prefix + i);
-      scope.scene.remove(axisLine);
-    });
+    var recursiveRemove = function(rootObj) {
+      if (rootObj.name != null && rootObj.name.startsWith(prefix)) {
+        scope.scene.remove(rootObj)
```
ElDeveloper:
```suggestion
        scope.scene.remove(rootObj);
```

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -591,21 +591,26 @@ define([
...
+      }
+      else {
+        // We can't iterate the children array while removing from it,
+        // So we make a shallow copy.
+        var childCopy = Array.from(rootObj.children)
```
ElDeveloper:
```suggestion
        var childCopy = Array.from(rootObj.children);
```

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -591,21 +591,26 @@ define([
...
+        for (child in childCopy)
+          recursiveRemove(childCopy[child])
```
ElDeveloper:
```suggestion
          recursiveRemove(childCopy[child]);
```

ElDeveloper:
Closing in favor of #755

PR #752, opened 2020-01-15
BLD: Drop support for Python 2.7

ElDeveloper:
:snake: 🔥 

We need to continue to use gjslint which only runs in 2.7.x (hence the invocation in .travis.yml) but otherwise we should be good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ matrix:
         env: NUMPY_VERSION="<1.13"
 
 before_install:
+  # run gjslint using py27
+  - sudo pip2.7 install https://github.com/google/closure-linter/archive/master.zip
   - wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh && chmod +x miniconda.sh
   - ./miniconda.sh -b
   - export PATH=/home/travis/miniconda2/bin:$PATH
@@ -31,8 +33,6 @@ before_install:
   - conda update --yes conda
 
 install:
-  # run gjslint using py27
-  - pip2.7 install https://github.com/google/closure-linter/archive/master.zip
   # install requests using conda to avoid a distutils error in certifi
   - conda create --yes -n travis python=$TRAVIS_PYTHON_VERSION pip numpy${NUMPY_VERSION} 'scipy>=0.17.0' matplotlib pandas flake8 pep8 jupyter coverage cython scikit-learn requests
   - source activate travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
 
 install:
   # gjslint only runs in py27
-  - conda create --yes -n gjslint python=2.7 pip
+  - conda create --yes -n gjslint python=2.7 pip six
   - source activate gjslint
   - pip install https://github.com/google/closure-linter/archive/master.zip
   - conda deactivate
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - 2.7
   - 3.5
   - 3.6
   # also includes 3.7 (see below)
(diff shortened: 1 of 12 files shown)


ElDeveloper:
**tl;dr** Python and JavaScript unit tests failures were not being reported in travis. I've fixed this in this PR, @antgonza and @dhakim87 can you both have a look?

The problem was that we were running the tests like:

```
coverage run tests/all_tests.py ; coverage report
```

Which meant that regardless of the output from the unit tests, the command `coverage report` would always succeed, thus shadowing the real output of the previous command. I changed this to:

```
coverage run tests/all_tests.py && coverage report
```

So now travis will be able to report if tests fail since the first command needs to succeed before the second command is executed.

emperor/support_files/js/ui-state.js
```diff
@@ -1,4 +1,4 @@
-define(['three', 'underscore'], function(THREE, _) {
+define(['three'], function(THREE) {
```
antgonza: Just to confirm, underscore is not used anymore, right?

.travis.yml
```diff
@@ -35,20 +31,29 @@ before_install:
...
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'; fi
   # we just check coverage in the latest version of NumPy
-  - coverage run tests/all_tests.py; coverage report
+  - coverage run tests/all_tests.py && coverage report
   - make -C doc html
+
+  # we can only run gjslint using py27
+  - conda deactivate
+  - conda activate gjslint
+  - gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'
```
thermokarst: You can collapse the two lines above into one:

```suggestion
  - conda run -n gjslint gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'
```

.travis.yml
```diff
@@ -35,20 +31,29 @@ before_install:
   - conda update --yes conda
 
 install:
+  # gjslint only runs in py27
+  - conda create --yes -n gjslint python=2.7 pip six
+  - source activate gjslint
+  - pip install https://github.com/google/closure-linter/archive/master.zip
```
thermokarst: You can collapse the line above and below this line into:

```suggestion
  - conda run -n gjslint pip install https://github.com/google/closure-linter/archive/master.zip
```

emperor/support_files/js/ui-state.js
ElDeveloper: You are correct, I just checked again and there's no matches to "_".

ElDeveloper:
Thanks for the reviews @thermokarst and @antgonza!

.travis.yml
```diff
@@ -35,20 +31,26 @@ before_install:
   - conda update --yes conda
 
 install:
+  # gjslint only runs in py27
+  - conda create --yes -n gjslint python=2.7 pip six
+  - conda run -n gjslint pip install https://github.com/google/closure-linter/archive/master.zip
+  - conda deactivate
```
thermokarst: This command shouldn't be necessary anymore, since the `conda run` above handles activation and deactivation, inline.

.travis.yml
```diff
@@ -35,20 +31,26 @@ before_install:
...
   - flake8 emperor/*.py tests/*.py scripts/*.py setup.py
-  # we can only run gjslint in a python 2.7.x environment
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'; fi
   # we just check coverage in the latest version of NumPy
-  - coverage run tests/all_tests.py; coverage report
+  - coverage run tests/all_tests.py && coverage report
   - make -C doc html
+
+  # we can only run gjslint using py27
+  - conda run -n gjslint gjslint --custom_jsdoc_tags 'module,function,constructs,alias,default' 'emperor/support_files/js/*.js' 'tests/javascript_tests/*.js'
```
thermokarst: Not sure how `conda run` behaves when executed inside another env - you might still need to `conda deactivate` on the line above.

.travis.yml
ElDeveloper: Thanks, removed!

.travis.yml
ElDeveloper: I checked locally and it turns out you can just use conda run directly ✨  Thanks for the suggestion!

thermokarst: Woot!

ElDeveloper:
Thanks so much both!

Issue #761, opened 2020-05-09
Inconsistent sorting for mixed-type (numeric + non-numeric) fields in the color legend


fedarko:
When a field contains mixed numeric and non-numeric values, the non-numeric values are correctly assigned colors from the "start" of the color palette: but they aren't placed in the legend before the numeric values. This results in the legend looking off (the red and blue colors are the first two in the `Classic QIIME Colors` map, but they aren't listed in this order in the screenshot below).

![image](https://user-images.githubusercontent.com/4177727/81459873-c239bc00-9156-11ea-9086-b047977bf879.png)

For reference, I'm attaching the sample metadata I used to generate the above visualization (it's just the Q2 moving pictures data), and the QZV file for the visualization (renamed to use .zip so that GitHub doesn't explode):

[sample_metadata_badscaling_test.txt](https://github.com/biocore/emperor/files/4602422/sample_metadata_badscaling_test.txt)

[unweighted_unifrac_emperor.zip](https://github.com/biocore/emperor/files/4602425/unweighted_unifrac_emperor.zip)

fedarko:
OK, I think I've isolated the source of the problem. It traces back to `DecompositionView.setCategory()`, which adds items to a `dataView` in what seems like an arbitrary order:

https://github.com/biocore/emperor/blob/023b6ecb761c31cd7f60a2e38e418d71199eb4e1/emperor/support_files/js/view.js#L769-L794

Flipping things around, so that we sort the `attributes`' keys first, seems to fix the problem:

```js
DecompositionView.prototype.setCategory = function(attributes,
                                                   setPlottableAttributes,
                                                   category) {
  var scope = this, dataView = [], plottables;

  var fieldValues = util.naturalSort(_.keys(attributes));

  _.each(fieldValues, function(fieldVal, index) {
    /*
     *
     * WARNING: This is mixing attributes of the view with the model ...
     * it's a bit of a gray area though.
     *
     **/
    plottables = scope.decomp.getPlottablesByMetadataCategoryValue(category,
                                                                   fieldVal);
    if (setPlottableAttributes !== null) {
      setPlottableAttributes(scope, attributes[fieldVal], plottables);
    }

    dataView.push({id: index, category: fieldVal, value: attributes[fieldVal],
                   plottables: plottables});
  });
  this.needsUpdate = true;

  return dataView;
};
```

This results in the legend elements being sorted as expected --

[...]

PR #763, opened 2020-05-11
Some (mostly coloring-focused) bug fixes

### LINKED ISSUE #760: "Incomplete" interpolation for sequential / diverging color maps
When selecting a sequential / diverging color map (e.g. Viridis), the interpolation is incomplete: the "end" color isn't the end of the color map. See the example below, where the end color should be #fee825 (yellow) but is actually #8cb373 (blue-ish green).

image

This notably doesn't seem to be a problem when using the "continuous values" scaling:

image

I suspect that [this line](https://github.com/biocore/emperor/blob/023b6ecb761c31cd7f60a2e38e418d71199eb4e1/emperor/support_files/js/color-view-controller.js#L468) in `color-view-controller.js` may be causing these problems, but I'm not sure.

### LINKED ISSUE #762: Update chroma.js version
As mentioned in https://github.com/biocore/empress/pull/159#issuecomment-625580999, the version used by Empress and Emperor is a few years out of date, so this results in tiny precision problems.

fedarko:
Closes #760 and closes #762. Also fixes that `naturalSort()` Infinity issue we talked about (there isn't an open issue for that though).

The JavaScript tests have ~~mostly~~ all been updated. This involved updating both the expected colors from the interpolation fixes, as well as updating slightly different colors due to the new Chroma.js version. (For the two tests that directly checked gradient SVG equality, I elected to just copy in the new expected SVG over the old expected SVG markup -- changing each hex color manually seemed kinda daunting, as shown in the QUnit diff below.)

errors

I also tested this manually, and can confirm that this resolves #760 -- see below, where the interpolated Viridis colors now extend to cover the full color map.

image


emperor/support_files/js/util.js
```diff
@@ -13,8 +13,12 @@ define(['underscore'], function(_) {
     var numericPart = [], alphaPart = [], result = [];
 
     // separate the numeric and the alpha elements of the array
+    // Note that NaN and +/- Infinity are not considered numeric elements
     for (var index = 0; index < list.length; index++) {
-      if (isNaN(parseFloat(list[index]))) {
+      if (
+          isNaN(parseFloat(list[index])) ||
```
wasade, line 19: Worth doing the parse to float a single time?

tests/javascript_tests/test_util.js
```diff
@@ -118,6 +118,47 @@ requirejs(['jquery', 'underscore', 'util'], function($, _, util) {
...
+     *
+     * Test that strings like "Infinity" are sorted with words in
+     * naturalSort's output, rather than with numbers. This test was adapted
+     * from Empress.
+     *
+     */
+    test("Test that naturalSort doesn't treat Infinity / NaN as numbers", function () {
+      var eles = [
+        "1",
+        "2",
```
wasade, line 131: is it important to include things like 1e10?

emperor/support_files/js/util.js
fedarko: yeah, would be nice (and should make this if statement fit on one line i guess). will resolve tomorrow

tests/javascript_tests/test_util.js
fedarko: I hadn't thought of that! Would be nice to have tested. That being said, it looks like in JS scientific notation stuff like 1e10 is interpreted correctly as a number by `parseFloat()`, so I think this should be handled as expected already (but ofc would be best to explicitly test for that :)

```js
> parseFloat("1e10");
10000000000
```

wasade: Excellent :)

fedarko:
Comments have been addressed. It's worth noting that scientific notation numbers like `1e2` are handled as expected by `naturalSort()` (and this is now explicitly tested for), _but_ the legend is still not sorting things correctly since this PR doesn't address #761. So datasets can still look something like

image

... where the colors are assigned correctly, but the legend is out of order.

If y'all would like me to I can try to address #761 in this PR as well, but I'm not sure where to look for that.

Thanks!

PR #765, opened 2020-05-23
Fix color sorting in the legend

### LINKED ISSUE #761: Inconsistent sorting for mixed-type (numeric + non-numeric) fields in the color legend

fedarko:
Closes #761; see https://github.com/biocore/emperor/issues/761#issuecomment-629555803 for details on the way I fixed the bug.

Along with the fix to `DecompositionView.setCategory()`, I also added in a test to make sure that these changes remain in effect. Also made some minor code fixes in `test_decomposition_view.js`.

Hope this is useful!


fedarko:
@ElDeveloper done, thanks!

fedarko:
For some reason, it looks like the gjslint / flake8 versions used on Travis changed between last week and this week -- hence why the build was failing. Things should be fixed now (most of the complaints were minor things like double vs. single quotes).

PR #766, opened 2020-05-26
Add brush selection

ElDeveloper:
The user can now hold shift and select a group of samples (see animated GIF):

brush-select

Now that the functionality is in place, we have to decide what do we do with the selected samples. For now the color is changed, and the sample list is printed to the console. Next, we can copy the list of samples to the clipboard. In any case this callback will be helpful to integrate with other visualization tools.

You'll notice this required that THREE was updated to the latest version. A number of API calls were updated along the way. Changes to any of those JavaScript files should pretty much be ignored.


antgonza:
Really cool! I guess a good start would be to copy the sample names to the clipboard so a user can copy/paste them to another tool. This, makes me wonder how this new feature will work in the large PCoA's we have been creating, for example the +113K samples one. Perhaps, then as a following step (other PR), it would be nice to be able to define some kind of API so that selection creates a button, that can be pressed and a new filter of the distance matrix and pcoa job can be sent do QIIME2/Qiita (I can imagine a config file that allows configuration of this feature) ...

ChangeLog.md
```diff
@@ -14,6 +14,16 @@ Emperor ChangeLog
   very slight precision differences in things like color interpolation
   ([#762](https://github.com/biocore/emperor/issues/762))
 
+### New features
+
+* Add sample selection support. Users can select samples by holding shift and
+  dragging the mouse. Selected samples are copied to the users' clipboard
```
antgonza, line 20: pointer? I guess mouse works ...

ChangeLog.md
```diff
@@ -14,6 +14,16 @@ Emperor ChangeLog
...
+  ([#153](https://github.com/biocore/emperor/issues/153)).
+
+### Miscellaneous
+
+* Updated THREE.js and plugins to the latest version (r116).
```
antgonza, line 25: 👍

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -899,6 +973,180 @@ define([
...
+   *
+   * Adds the mouse selection events to the current view
+   *
+   * @param {node} $container The container to add the events to.
+   * @private
+   */
+  ScenePlotView3D.prototype._addSelectionEvents = function($container) {
+    var scope = this;
+
+    // There's three stages to the mouse selection:
```
antgonza, line 1050:
```suggestion
    // There're three stages to the mouse selection:
```

Anyway, really cool!

emperor/support_files/js/sceneplotview3d.js
```diff
@@ -899,6 +973,180 @@ define([
...
+    // The mousdown event is ignored unless the user is holding Shift. Once
+    // selection has started the rotation controls are disabled. The mousemove
+    // event continues until the user releases the mouse. Once this happens
+    // rotation is re-enabled and the selection box disappears. Selected
+    // markers are highlighted by changing the light they emit.
+    //
+    $container.on('mousedown', function(event) {
+      // ignore the selection event if shift is not being held or if parallel
+      // plots are being visualized at the moment
+      if (!event.shiftKey || scope.parallelController.enabled) {
```
antgonza, line 1062: it would be cool to allow this in parallel plots .... just a thought

ElDeveloper: I looked at this and it wasn't much harder to add support for parallel plots so this now works:

parallel-plot

ElDeveloper:
>  Only a few minor comments; BTW should all three.js-plugins/*.js be minimized?

We haven't in the past since there's a couple of things that we modify for our own code. Some patches which have been accepted in mrdoob/three.js some which haven't (yet).

ElDeveloper:
Thanks for reviewing @antgonza!

PR #767, opened 2020-05-30
Add "figure mode" button [WIP]

### LINKED ISSUE #604: 2d sample glyphs for 2d mode
It looks a little strange when exporting a 2d image to see 3d objects. (It turns out the cube bug from #601 actually work pretty well for this purpose).

It would be nice to have 2d sample glyphs for publications.

### LINKED ISSUE #753: Adding a "figure mode" button
This would be a single button accessible in the UI that does the following things:

- changes the plot to 2D
- changes the background to white
- changes sample shapes to 2D circles (#607)
- ideally, would move axis labels off of the actual axes (this might be worth a separate issue)

Eventually I'd like to take a shot at this, but don't really have time right now... I think this would make a lot of people happy.

(Came up during discussion with some folks in the lab.)

fedarko:
Closes #753 and closes #604.

This PR adds a small lightbulb icon in the top-right corner of the display, directly to the left of the "More Options" button:

image

On clicking this button, the plot changes to use more easily-displayable-in-a-paper settings:

image

The exact things that this button does are:

1. Make the plot 2D
2. Change the axis color to black
3. Change the background color to white
4. Change all of the shapes to circles (#604)

The first three of these things required minimal extra code (as is shown in `controller.js`, each takes up one line of code -- I attribute this to Emperor's APIs being really well designed :).

The fourth thing (making everything a circle), however, required more work -- and this is what I'm most uncertain about here.

This PR adds "Circle" as a selectable shape, and it seems to work pretty well in the JS interface -- but there are some artifacts in the exported screenshots, for example with the Jaccard moving pictures PCoA:

_In-Emperor_
image

_Exported PNG_ (note the "triangle" borders within each circle, as well as the slight inconsistency of that one blue dot at the top being partially on top of orange)
emperor (2)

_Exported PNG (zoomed in at the top)_
image

Anyway, I was thinking we may we want to be smarter about shapes for the 2D mode anyway --  and do something similar to how `DecompositionView._buildVegaSpec()` [has an object set up to convert 3D shapes to "close" 2D shapes](https://github.com/biocore/emperor/blob/e874f0416104449fe32fc1a50b8c517369623269/emperor/support_files/js/view.js#L1086-L1116). That might require extra work due to a need to add more 2D shapes to Emperor, so before doing that I wanted to check to see what folks would prefer (we could also conceivably not change the shapes at all, and/or leave doing that to a separate PR). So this is marked as a draft for now.

(Also, this PR should have tests, at least for things like `makeEverythingACircle()`.)

### DIFF
--- a/emperor/support_files/js/controller.js
+++ b/emperor/support_files/js/controller.js
@@ -266,6 +266,24 @@ define([
     });
     this.$plotSpace.append(this.$optionsButton);
 
+    /**
+     * @type {Node}
+     * jQuery object
+     */
+    this.$displayModeButton = $('<button name="dm-button">&nbsp;</button>');
+    this.$displayModeButton.css({
+      'position': 'absolute',
+      'z-index': '3',
+      'top': '5px',
+      'right': '40px'
+    }).attr('title', 'Display Mode').on('click', function(event) {
+        scope.controllers.axes.updateVisibleAxes(null, 2);
+        scope.controllers.axes._colorChanged("axes-color", "#000000");
+        scope.controllers.axes._colorChanged("background-color", "#ffffff");
+        // TODO: show circles instead of spheres; fix axis labels
+    });
+    this.$plotSpace.append(this.$displayModeButton);
+
     // default decomposition view uses the full window
     this.addSceneView();
 
@@ -273,6 +291,9 @@ define([
       // setup the jquery properties of the button
       scope.$optionsButton.button({text: false,
                                    icons: {primary: ' ui-icon-gear'}});
+
+      scope.$displayModeButton.button({text: false,
+                                   icons: {primary: ' ui-icon-star'}});
 
       scope._buildUI();
       // Hide the loading splashscreen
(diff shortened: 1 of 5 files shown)


emperor/support_files/js/controller.js
```diff
@@ -266,6 +266,25 @@ define([
...
+     * jQuery object
+     */
+    this.$figureModeButton = $('<button name="dm-button">&nbsp;</button>');
+    this.$figureModeButton.css({
+      'position': 'absolute',
+      'z-index': '3',
+      'top': '5px',
+      'right': '40px'
+    }).attr('title', 'Figure Mode').on('click', function(event) {
+        scope.controllers.axes.updateVisibleAxes(null, 2);
```
ElDeveloper, line 280: For updating different attributes of the axes controller, I would suggest using `scope.controllers.axes.fromJSON`, then you encode these attributes as an object and you should be good to go (see the unit tests for some examples).

emperor/support_files/js/shape-controller.js
```diff
@@ -106,6 +106,30 @@ define([
...
+   *
+   * NOTE: In the future, making this pay more close attention to the current
+   * shapes of each object (like how DecompositionView._buildVegaSpec() works)
+   * may be preferable. However, that may require adding more 2D shape options
+   * to Emperor.
+   *
+   * @extends EmperorAttributeABC
+   *
+   */
+  ShapeController.prototype.makeEverythingACircle = function() {
```
ElDeveloper, line 121: Can you add a unit test for this method?

PR #769, opened 2020-06-16
Fix bug that would fail to load axes settings

ElDeveloper:


emperor/core.py
```diff
@@ -1278,6 +1282,10 @@ def set_axes(self, visible=None, invert=None, color='white'):
             raise ValueError('One or more of your visible dimensions are out '
                              'of range.')
 
+        if view_type not in {'scatter', 'parallel-plot'}:
+            raise ValueError('Unknown vew type options are "scatter" or '
```
antgonza, line 1286:
```suggestion
            raise ValueError('Unknown view type options are "scatter" or '
```

ElDeveloper:
Thanks @antgonza added!

ElDeveloper:
Thanks @antgonza!

Issue #771, opened 2020-06-17
Selection of parallel plot markers yield the shader object


ElDeveloper:
While this is fine within the context of Emperor, it makes the sample selection not work without a lookup table fo the plottables (related to https://github.com/biocore/empress/pull/174).

Issue #747, opened 2020-01-08
Parallel plot display looks odd when activated from a biplot, and causes issues in the scatterplot view


fedarko:
Just ran into this today -- figured I should raise an issue here.

## Steps to reproduce

1. Open up a biplot in Emperor. I'm using the biplot produced in [this notebook](https://nbviewer.jupyter.org/github/biocore/qurro/blob/master/example_notebooks/DEICODE_sleep_apnea/deicode_example.ipynb) with [this version of DEICODE](https://github.com/biocore/DEICODE/pull/54), for reference.

<img width="1435" alt="Screen Shot 2020-01-07 at 4 28 24 PM" src="https://user-images.githubusercontent.com/4177727/71940371-ef576d00-316a-11ea-8c75-e037e2437173.png">

2. Switch to `Parallel Plot`.
**Problem 1:** first this shows a zoomed-in view of the parallel plot, which seems odd but not a huge problem.

<img width="1437" alt="Screen Shot 2020-01-07 at 4 30 03 PM" src="https://user-images.githubusercontent.com/4177727/71940403-0d24d200-316b-11ea-8810-8b4969143837.png">

**Problem 2:** However, when you zoom out in the parallel plot view you also see that some of the arrows from the biplot are still hanging around. (My guess is that the presence of these arrows is messing with the camera somehow, but I can't confirm that.)

<img width="1440" alt="Screen Shot 2020-01-07 at 4 30 55 PM" src="https://user-images.githubusercontent.com/4177727/71940445-30e81800-316b-11ea-9601-c1fb85a88c30.png">

3. Switch back from `Parallel Plot` to `Scatter`.
**Problem 3:** For some reason, one of the 3 arrows from the biplot is now missing?

<img width="1439" alt="Screen Shot 2020-01-07 at 4 36 50 PM" src="https://user-images.githubusercontent.com/4177727/71940656-edda7480-316b-11ea-8445-02a30f761d16.png">

## Desired behavior
Ideally:
- The camera should be initialized to show a complete view of the parallel plot
- The parallel plot should not contain the arrows 
- All displayed arrows should be preserved when moving back to the biplot

## Workaround
In practice, if parallel plots will cause lots of problems with biplots, I guess it might make sense to just temporarily disable parallel plots for biplot visualizations -- this way at least users won't run into the above problems (in particular problem 3, which IMO is the most scary). But it's not really my call :)

Thanks!

ElDeveloper:
[...]

PR #773, opened 2020-06-22
Fixes an issue with parallel plots and augments the select API

ElDeveloper:
- When a user visualized biplots, they wouldn't be able to swtich to parallel plot mode without some markers being left behind. This needed a bit of refactoring of some underlying methods #747.

- Expand the selection API to return the view object (this is for better integration with Empress) - #771 .

### DIFF
--- a/emperor/support_files/js/sceneplotview3d.js
+++ b/emperor/support_files/js/sceneplotview3d.js
@@ -176,6 +176,7 @@ define([
         scope.scatterController.enabled = false;
         scope.parallelController.enabled = true;
         scope._selectionBox.camera = scope.camera;
+        scope._selectionBox.collection = [];
       } else {
         scope.camera = scope.scatterCam;
         scope.control = scope.scatterController;
@@ -183,6 +184,7 @@ define([
         scope.scatterController.enabled = true;
         scope.parallelController.enabled = false;
         scope._selectionBox.camera = scope.camera;
+        scope._selectionBox.collection = [];
       }
       //Disable any active rotation
       if (evt.newVal === 'parallel-plot')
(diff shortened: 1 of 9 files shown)


emperor/support_files/js/sceneplotview3d.js
```diff
@@ -324,41 +326,7 @@ define([
       copyToClipboard(n);
       showText('(copied to clipboard) ' + n, i);
     });
-    this.on('select', function(selected) {
```
ElDeveloper, line 327: The code here was moved down in the file.

emperor/support_files/js/view.js
```diff
@@ -1097,6 +1104,106 @@ DecompositionView.prototype.toggleLabelVisibility = function() {
   this.needsUpdate = true;
 };
 
+
+/**
+ * Set the emissive attribute of the markers
+ *
+ * @param {Bool} emissive Whether the object should be emissive.
+ * @param {Plottable[]} group An array of plottables for which the emissive
+ * attribute. If this object is not provided, all the plottables in the
```
antgonza, line 1113:
```suggestion
 * attribute will be set. If this object is not provided, all the plottables in the
```

emperor/support_files/js/view.js
```diff
@@ -1097,6 +1104,106 @@ DecompositionView.prototype.toggleLabelVisibility = function() {
...
+ * view will be have the scale set.
+ */
+DecompositionView.prototype.setEmissive = function(emissive, group) {
+  group = group || this.decomp.plottable;
+
+  if (this.decomp.isArrowType()) {
+    throw new Error('Cannot set emissive attribute of arrows');
+  }
+
+  var i = 0, j = 0;
```
antgonza, line 1123: Is this OK? I mean, these variables are used in the for's below ...

emperor/support_files/js/view.js
```diff
@@ -1097,6 +1104,106 @@ DecompositionView.prototype.toggleLabelVisibility = function() {
...
+
+  if (this.UIState.getProperty('view.usesPointCloud') ||
+      this.UIState.getProperty('view.viewType') === 'parallel-plot') {
+    var emissives = this.markers[0].geometry.attributes.emissive;
+
+    // the emissive attribute is a boolean one
+    emissive = (emissive > 0) * 1;
+
+    if (this.markers[0].isPoints) {
+      for (i = 0; i < group.length; i++) {
```
antgonza, line 1133: ... perhaps related to my previous question, should this for's be `for (var i = 0 ...`

emperor/support_files/js/view.js
ElDeveloper: Yes, I ended up doing this because we have a couple `for` loops where I would have needed to use `var i` and the linter would have complained that I was declaring each variable multiple times. The solution to this issue is to declare all of these at the top of the scope.

emperor/support_files/js/view.js
ElDeveloper: Same as above, it's to avoid he linter from complaining about variables being re-declared.

ElDeveloper:
Thanks for the review @antgonza!

PR #783, opened 2020-08-26
Fixing scatterplot

mortonjt:
This function will currently break due to changes in the skbio API -- OrdinationResults expects a pandas series as input for the eigvals object.


ElDeveloper:
Thanks @mortonjt!

PR #789, opened 2020-10-14
Force gradient SVG to cover the full color map and gradient ranges

### LINKED ISSUE #788: Gradient SVG is slightly inaccurate
The actual coloring of samples is fine, I think, but the SVG gradients produced _sometimes_ cover only ~99% of the domain, not 100%. This means that the end colors shown in the gradients are very slightly incorrect.

It's pretty confusing, because right now there are two tests that test the gradient SVG creation:

1. In [one of these tests](https://github.com/biocore/emperor/blob/b625a5403ce9ad0ef3a84c6bc0b68b33e97888a4/tests/javascript_tests/test_color_view_controller.js#L212-L325) this error is present. The returned Viridis gradient, which is used as the reference for verifying the result, ends at 99% (with `#f8e725`), rather than at 100% (`#fee825`, which is the actual end color for Viridis).

2. In [the other one of these tests](https://github.com/biocore/emperor/blob/b625a5403ce9ad0ef3a84c6bc0b68b33e97888a4/tests/javascript_tests/test_color_view_controller.js#L348-L471), this error does not seem to be present: the color map being tested is Blues, and the minimum and maximum colors (`#f7fbff` and `#08306b`) match up with the extrema for this color map (going off what `chroma.scale('blues')(0)` and `chroma.scale('blues')(1)` return).

I think the problematic part of code is within this block:

https://github.com/biocore/emperor/blob/b625a5403ce9ad0ef3a84c6bc0b68b33e97888a4/emperor/support_files/js/color-view-controller.js#L426-L439

The first for loop seems suspicious to me. I'm not completely sure yet what's wrong with this code, but it looks like there might be precision issues with the `s <= max` check: since the data used for the test that failed had a `step` of `(4 - 0) / 100 = 0.04`, while the data used for the test that succeeded had a relatively larger `step` of `(50 - 0) / 100 = 0.5`.

I think it would make the most sense to just remove some of this code and use [`chroma.scale.colors()`](https://gka.github.io/chroma.js/#scale-colors) to do the work for us:

```js
 var stopColors = interpolator.colors(101);
// Now we can construct gradientSVG just like we did before, with the final color (corresponding
// to the maximum value) being placed at 100%
```

Can put together a PR for this tomorrow if you'd like.

This impacts Empress as well, since the gradient SVG code there was adapted from this code.

fedarko:
Closes #788. This solution should also be applicable to https://github.com/biocore/empress/issues/421.


fedarko:
Not sure why the build is failing -- it doesn't look like it's the JS tests' problem. Seems like it might be a NumPy version issue, judging by the errors showing up and the fact that just the `NUMPY_VERSION="<1.13"` builds fail.

image

ElDeveloper:
@fedarko, would you be able to `git pull upstream master` to get the latest changes in your local branch. And then push to re-run the CI. I just migrated the project from Travis to GitHub actions so things should *work*. 😟

fedarko:
@ElDeveloper just re-pushed; looks like the build is passing now.

ElDeveloper:
Thanks so much @fedarko. One last request, can you add a note to the ChangeLog? Other than that, this should be ready to go.

fedarko:
@ElDeveloper -- Got it, added a note to the changelog. Thanks!

ElDeveloper:
Thanks @fedarko !

PR #793, opened 2021-02-09
Added a new "info" attribute

ahdilmore:
This attribute is displayed in the top left of the plot's banner. It is useful for displaying statistics such as the M^2 value and p value of a procrustes analysis, which QIIME2 previously had no way of outputting or visualizing (discussed [here](https://github.com/qiime2/q2-diversity/issues/210)). An example display is shown in the image below: 
image


ElDeveloper (approved):
Looks great, thanks @ahdilmore. Only missing piece is the changelog note.

PR #808, opened 2022-01-27
BUG: Remove rawgit and a bit of jQuery

### LINKED ISSUE #751: RawGit has reached the end of its useful life
Plots that are generated with the argument `remote=True` will not work anymore.

ElDeveloper:
- We use jsdelivr instead of rawgit which has better support for
  versioning, etc.

- The style template needed jQuery but some distributions, such as Voilá,
  would not make jQuery available making this template crash.

Fixes #751

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,12 +5,15 @@ Emperor ChangeLog
 
 ### Bug Fixes
 
+* Update remote URL to no longer use rawgit.
+  ([#751(https://github.com/biocore/emperor/issues/751)]).
+
 ### New Features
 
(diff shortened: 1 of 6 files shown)


antgonza:
Thank you for submitting these changes, they look fine - just to confirm, is it OK to merge?

ElDeveloper:
Should be good to go, same as the other open PR 👍

Yoshiki Vázquez-Baeza

ElDeveloper:
Thanks @antgonza!

Yoshiki Vázquez-Baeza

PR #830, opened 2026-01-12
replace deprecated assertRaisesRegexp with assertRaisesRegex

a-detiste:
Hi,

This was a second patch that has been laying around in the Debian packaging for a long time:

Same reasoning as here: https://github.com/biopython/biopython/issues/2273


ElDeveloper (approved):

ElDeveloper:
Thanks @a-detiste

